PR #2, opened 2015-03-24
Fix download links for apache-maven-{3.2.5,3.1.1,3.0.5}-src.tar.gz.asc.

Stephan202:
Summary says it. Some links on the download page contain the `/maven-2/` path component, while that should be `/maven-3/`.


hboutemy:
patch applied, thank you

Stephan202:
Thanks, will close the PR then :).

PR #3, opened 2015-08-09
Link to the matrix

jbaruch:
Re-added a link to the matrix back. It's a good and important source of information.

### DIFF
--- a/content/markdown/repository-management.md
+++ b/content/markdown/repository-management.md
@@ -40,3 +40,4 @@ and the features provided by these products.
 * <a href="http://www.sonatype.org/nexus/go/" target="_blank" rel="nofollow">Sonatype Nexus OSS</a> (open source)
 * <a href="http://links.sonatype.com/products/nexus/pro/home" target="_blank" rel="nofollow">Sonatype Nexus Pro</a> (commercial)
 
+[Continuously updated by the maintainers of Archiva, Artifactory and Sonatype feature comparison matrix.](https://binary-repositories-comparison.github.io/)


mosabua:
We do not want to link to any external source for comparison that is not managed by all stakeholders and we do not want to manage to be an impartial source. We therefore just link to the different pages of the different products. The rest is up to the user.

mosabua:
And as such -1 ... do not merge imho.

jbaruch:
well, not that I am surprised. Regarding "not managed by all stakeholders" - this is a lie. @olamy and @brianf (the authors of the original matrix) are invited admins of the comparison matrix github repo.

mosabua:
Neither @brianf nor @olamy have interest nor time from my conversations with them to make sure the matrix stays independent and neutral. As such it is NOT impartial.  E.g. just look at the commit history.

https://github.com/binary-repositories-comparison/binary-repositories-comparison.github.io/commits/master

Feel free to put the matrix on the jfrog website and link from there, but the Maven project should imho stay out of this political tit for tat.

Nothing stops you from linking to the matrix from the jfrog site or any other website.

jbaruch:
It takes the exact amount of time and effort as the previous matrix on
codehaus.
Re the commit history, that's ridiculous claim. The migration from codehaus
took some commits, obviously. And the fact that Archiva and Nexus
shareholders didn't update their columns is not because they can't. They
did it for years and I can't see a single reason why they'd want to stop.
Just merge this PR and see how they both update the content.

stephenc:
As a member of the PMC:

The PMC has discussed this. The decision of the PMC is that the Apache Maven site will present an alphabetically sorted list of repository managers only.

PR #6, opened 2016-09-01
Enumerate inherited POM elements more thoroughly

ctrueden:
The documentation gave the impression that only a small subset of elements are inherited, but the truth is that most of them are. It is useful here also to list elements which are _not_ inherited, so that developers are fully informed of what to expect regarding inheritance.


ctrueden:
@khmarbaise Hopefully this makes sense!

mfriedenhagen:
@ctrueden, are you sure name is not inherited? Description is unfortunately.

ctrueden:
Yep, in order to make that list, I did a test: `<name>` is indeed not inherited. I was surprised, but in retrospect, it makes sense.

ctrueden:
Any obstacles to merging this?

ctrueden:
Is there some other way which would be better for me to submit patches?

mfriedenhagen:
Sorry for the late response, as agreed on the dev list, no JIRA issue is needed. I integrated your changes in http://svn.apache.org/viewvc?view=revision&revision=1777914.

ctrueden:
Thank you @mfriedenhagen!

PR #7, opened 2016-09-28
Corrected assumption regarding execution's configuration

fmarot:
Corrected a wrong assumption saying that specific plugin's execution's configuration could not be used from the command line.
This used to be true but not since Maven 3.3.1 and the implementation of MNG-5768 ( https://issues.apache.org/jira/browse/MNG-5768 ).
From the command-line, the use of the specific syntax with "@" specifying a specific execution (instead of "default-cli) will make the plugin's execution's specific <configuration> section used.

### DIFF
--- a/content/apt/guides/mini/guide-configuring-plugins.apt
+++ b/content/apt/guides/mini/guide-configuring-plugins.apt
@@ -570,13 +570,21 @@ public class MyBindedQueryMojo
 
   Now, <<<MyBindedQueryMojo>>> default phase which is package has been overrided by
   install phase.
-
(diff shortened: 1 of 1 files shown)


ssisk:
I had questions about this very topic while reading the docs, and this commit would have been very helpful in helping me understand it.  Thanks for writing it up! I hope it can be merged soon.

fmarot:
Hi @pono , may I ask why this issue is closed ? This is something I struggled with and corrected the doc to help people having the same questions that I answered. It also seems like it would have helped @ssisk . So why close it without merging it, or at least improving the doc if my proposal was not good quality ?

PR #22, opened 2018-07-08
Fix an extraneous word

AnAverageHuman:


khmarbaise:
Thanks for finding this and contributing this fix...

PR #27, opened 2018-08-19
Remove redundant 'that' in Minimal POM paragraph

johananl:
The following sentence doesn't make sense at the English level:

"Furthermore, as you can see that in the minimal POM, the repositories
were not specified."

I'm guessing the author has meant one of the following:

"Furthermore, as you can see in the minimal POM, the repositories were
not specified."

"Furthermore, you can see that in the minimal POM the repositories
were not specified."

I've chosen the second option because in my opinion it is clearer.


johananl:
I am not a native English speaker either, however sometimes mistakes pop in front of my eyes, and when they do - I do my best to fix them and contribute whenever I can :-)

Thank you for the fast feedback!

khmarbaise:
So welcome to the club ;-) Nevertheless thanks for finding this...and of course if you see something do not hesitate to contribute.

PR #33, opened 2018-08-29
Remove some Maven 1 history.

elharo:
This was over a decade ago. Let's just specify what happens today.


jira-importer:
Resolve #1095

PR #36, opened 2018-09-08
Fixed 2 links

dc55028:
an "i" was missing in the CMD and Checkstyle maven plugin links


khmarbaise:
Thanks for your contribution. Will be life in a few minutes...

PR #39, opened 2018-09-21
Fix broken links using archive.org

jdambron:
Some old links are broken, we could use archive.org to still be able to see the content of the articles.


hboutemy:
great idea, thank you

PR #40, opened 2018-09-26
More accurate description of artifacts

elharo:


hboutemy:
notice that an artifact is not identified by a packaging: it is identified by a classifier and an extension, which can eventually be named by a type
see https://maven.apache.org/ref/3.5.4/maven-core/artifact-handlers.html for the list of default types and corresponding extension + classifier
and POM dependency description: https://maven.apache.org/ref/3.5.4/maven-model/maven.html#class_dependency

elharo:
Thanks. Part of what I'm trying to do here is get a clear understanding of all these pieces, and then make sure what we say is accurate and consistent everywhere. On http://maven.apache.org/pom.html it says 

We will visit the classifier later, but for now it suffices to know that these kinds of projects are displayed as groupId:artifactId:packaging:classifier:version.

Is that incorrect? Or si something more subtle going on here where one coordinate like  groupId:artifactId:packaging:classifier:version identifies multiple artifacts of different types? In that case what do we call the thing indicated by groupId:artifactId:packaging:classifier:version if not artifact?

hboutemy:
thanks for your help.
"groupId:artifactId:packaging:classifier:version" is definitely wrong: it's "groupId:artifactId:extension:classifier:version"
packaging cannot become part of coordinates: it's a build lifecycle, and during a build, multiple artifacts will be created, each with different extension and classifier. And if you look at artifact handlers, https://maven.apache.org/ref/3.5.4/maven-core/artifact-handlers.html , you'll see that the packaging for java-source and javadoc does simply not mean anything. In POM dependency documentation, we wrote "The type often corresponds to the packaging used, though this is also not always the case."

PR #42, opened 2018-09-27
Distinguish packaging and extension

elharo:
@hboutemy packaging in POM; extension in coordinates

Please review carefully. I likely don't have this quite right.

### DIFF
--- a/content/apt/pom.apt
+++ b/content/apt/pom.apt
@@ -201,7 +201,7 @@ POM Reference
 
 * {Maven Coordinates}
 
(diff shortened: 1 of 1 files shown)


hboutemy:
I took the good ideas you had in this PR but reworked more in depth the paragraph, since packaging is definitely not part of coordinates: I hope the new paragraph is more clear
thank you for your help, I knew for a long time that there was some old documentation to fix, but never had energy to work on it with the depth required before you helped me a lot

jira-importer:
Resolve #1095

PR #43, opened 2018-10-02
Rewrite intro paragraph

elharo:
This one's optional. There was nothing incorrect in the original, but I thought we might be clearer, more active, and somewhat more specific about the benefits of Maven.


elharo:
Ping. Any thoughts?

hboutemy:
I like it, merged
thank you

PR #49, opened 2018-11-05
fixed the URL referred to in the documentation

pradeepanchan:
The repo url did not match that mentioned in the POM.xml

### DIFF
--- a/content/apt/guides/introduction/introduction-to-the-pom.apt
+++ b/content/apt/guides/introduction/introduction-to-the-pom.apt
@@ -131,7 +131,7 @@ Introduction to the POM
 
  Furthermore, you can see that in the minimal POM the <repositories> were not specified. If you build your project using the minimal POM,
  it would inherit the <repositories> configuration in the Super POM. Therefore when Maven sees the dependencies in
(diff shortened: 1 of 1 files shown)


hboutemy:
where are you looking at?
see https://maven.apache.org/ref/3.6.0/maven-model-builder/super-pom.html or https://maven.apache.org/ref/3.0.5/maven-model-builder/super-pom.html

pradeepanchan:
I noticed it here: https://maven.apache.org/guides/introduction/introduction-to-the-pom.html

hboutemy:
ok, I see: this section of introduction to the pom should be updated with Maven 3.x version of super-pom
these old versions for Maven 2.0 or 2.1 are really obsolete now

slachiewicz:
Thank you @pradeepanchan for finding inconsistencies in the links.
I have made changes in 53798cc27aea3c67e4a2ece87e9d0e7fc0e47d4e, so that I use the newest pom from 3.5.4. Later versions did not make any changes to this file.
We can close this PR because it is no longer valid. 
I invite you to review and make comments and corrections that are always welcome. We apologize for waiting for a response.

PR #50, opened 2018-11-05
added headers to the tables to be more specific

pradeepanchan:
The left column was not clear on whether it refers to a phase or goal. I understood after following a stackoverflow Q&A. So i thought its better to update the documentation.


hboutemy:
good idea, merged
thank you

PR #51, opened 2018-11-05
updated incorrect information

pradeepanchan:
plugin is mistakenly referred to as a goal

### DIFF
--- a/content/apt/guides/mini/guide-configuring-plugins.apt
+++ b/content/apt/guides/mini/guide-configuring-plugins.apt
@@ -187,10 +187,10 @@ mvn myquery:query -Dquery.url=http://maven.apache.org
 ** {Help Goal}
 
   Recent Maven plugins have generally an <<<help>>> goal to have in the command line the description of the plugin,
-  with their parameters and types. For instance, to understand the javadoc plugin, you need to call:
+  with their parameters and types. For instance, to understand how to use the <<<javadoc>>> plugin, you need to call:
 
 +----+
-mvn javadoc:help -Ddetail -Dgoal=javadoc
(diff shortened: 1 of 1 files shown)


hboutemy:
I think your remark is true, but the solution introduces another mistake: I think we should talk about "the javadoc goal of the javadoc plugin", isn't it?

pradeepanchan:
Yes, you are right. When i tried it rephrase it, it seemed to make it more confusing, because javadoc appears as a goal as well as a plugin in the command. So i changed the command itself. I hope it looks more clear now.

hboutemy:
perhaps showing plugin description then goal details would be the best for an intro: in fact, the help goal can do both, and the hard part is for goal details

content/apt/guides/mini/guide-configuring-plugins.apt
```diff
@@ -187,10 +187,10 @@ mvn myquery:query -Dquery.url=http://maven.apache.org
 ** {Help Goal}
 
   Recent Maven plugins have generally an <<<help>>> goal to have in the command line the description of the plugin,
-  with their parameters and types. For instance, to understand the javadoc goal, you need to call:
+  with their parameters and types. For instance, to understand how to use the <<<javadoc>>> plugin, you need to call:
 
 +----+
-mvn javadoc:help -Ddetail -Dgoal=javadoc
+mvn javadoc:help
```
elharo: This command is not the same as the one on the left. I'm not sure which is needed here. They provide different info. The one on the left does seem to be for the goal while this one is for the plugin.   So I think the existing text is correct, though perhaps this is too. They're just talking about different things.

PR #57, opened 2018-12-09
Add java9+ workaround: use new surefire version

ecki:
As discussed on maven-users till quickstart archetype 1.4 (MARCHETYPES-63) is ready.

### DIFF
--- a/content/apt/guides/getting-started/maven-in-five-minutes.apt
+++ b/content/apt/guides/getting-started/maven-in-five-minutes.apt
@@ -140,6 +140,12 @@ my-app
 </project>
 +-----+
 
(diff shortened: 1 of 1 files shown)


ecki:
BTW: the site build also prints a warning about missing version and fails with Java 11, should we describe that also or just fix it in 1.4?

Add:
		<plugin>
          <groupId>org.apache.maven.plugins</groupId>
          <artifactId>maven-project-info-reports-plugin</artifactId>
          <version>2.7</version>
        </plugin>

hboutemy:
thank you for the proposition: I reworked it (fixing some details, like the issue with Java 11 but not Java 9) and published the result

notice: I just started the vote for quickstart archetype 1.4 with everything fixed...

PR #66, opened 2019-01-29
Add missing word and tighten

elharo:
@hboutemy FYI, I'm tempted to delete this note completely since it describes a situation that doesn't happen (compile+compile --> runtime), and I'm not sure a typical user of Maven needs to think about this. They just need to know what Maven does do, not what we sort of wish it could do.


hboutemy:
thank you

PR #68, opened 2019-01-30
Modernize

elharo:
@hboutemy This page hasn't been touched in a while. I made some obvious updates and removed or fixed broken links. However, I'm not sure if anything here is still accurate or not.


hboutemy:
thank you for these improvements: I'm not sure many people do https authentication, but it's useful to keep this documentation as good as possible

PR #69, opened 2019-02-05
Maven does not provide mailing lists

elharo:
@hboutemy Rather it can list the location of existing mailing lists, but it in no way creates or manages them. This is different than Cross referenced sources or change log documents, which it actually does create.


hboutemy:
you're right, Maven provide *the location* of mailing lists

hboutemy:
fixed in 51fade11eacd09f70f2aefac666f1a5ea10ee20a

PR #70, opened 2019-02-10
subject verb agreement

elharo:
@hboutemy


hboutemy:
good catch: thank you

PR #72, opened 2019-02-18
You can upload artifacts that depend on unreleased artifacts

elharo:
@hboutemy I know from brutal experience that it is possible to upload artifacts that depend on artifacts not yet released.

### DIFF
--- a/content/apt/repository/guide-central-repository-upload.apt
+++ b/content/apt/repository/guide-central-repository-upload.apt
@@ -32,13 +32,13 @@ Guide to uploading artifacts to the Central Repository
  In order for users of Maven to utilize artifacts produced by your project, you must deploy them to
  a remote repository. Many open source projects want to allow users of their projects who build with
  Maven to have transparent access to their project's artifacts. In order to allow for this, a project
- must have their artifacts deployed to {{{/repository/}the Central Repository}}.
+ must deploy their artifacts to {{{/repository/}the Central Repository}}.
 
 
 Requirements
 
- Only <releases> can be uploaded to the central repository, that means files that won't change and that only depend on other files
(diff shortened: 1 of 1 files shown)


hboutemy:
@elharo this is why you are removing the "and that only depend on other files already released and available" sentence part?

question: by "not yet released", you mean that you could upload:
1. to the Central Repository
2. SNAPSHOT versions or release versions not published to the Central Repository

notice: while looking at your proposal, I think it will be important to name precisely "the Central Repository", as it has specific rules and controls that don't happen on any other repository

than does not mean that these controls are perfect: I know that you cloud have a dependency on a release that is not available in the Central Repository because it may be available in another repository. That's not a good practice, there are some dreams to remove this flexibility for better safety, but flexibility vs safety is always a hard compromise...

elharo:
It's a little worse than that. You can (or at least you could six months ago)  publish com.example:foo:1.27 that depends on com.example:bar:1.27 that doesn't exist anywhere without uploading com.example:bar:1.27 first. You can't build anything that depends on com.example:foo:1.27, and this immediately breaks everyone who used a version range to pull in com.example:foo. 

https://github.com/GoogleCloudPlatform/appengine-gcs-client/issues/72

hboutemy:
yes, what you give as example is just a consequence of not really checking that a dependency is available perhaps in a different repository
you just add that with version ranges, the impact is terrific: you see why I hate version ranges for dependencies, and if it were my own choice, I would remove this feature :)

if we go back to your PR, I'm ok with the first and last improvements, but not with the removal of the requirement that dependencies must be available: even if this requirement is not strictly checked when uploading, the requirement is there and should really be explained rather than removed

PR #76, opened 2019-03-29
Cleanups including more white space between section in examples

elharo:
@hboutemy I'm still having trouble following this page, which is why I keep coming back to it. I'm thinking about sending a separate that modifies the group ID and artifact IDs in the examples and uses those for the references.


elharo:
Ping

PR #88, opened 2019-06-11
Adding extension to support packaging type maven-archetype in creating archetypes guide

aviemzur:
Using the pom in creating archetype guide results in the following error:
```
[ERROR]     Unknown packaging: maven-archetype @ line 9, column 16
```
I was able to resolve this by adding `archetype-packaging` extension.

Suggesting this change to the page to supply users with a working version of the example archetype pom.

Please correct me if there is a different way to handle this error, I found no other way.


aviemzur:
@hboutemy can you please tag a relevant reviewer?

hboutemy:
great finding: thank you

PR #98, opened 2019-08-26
Improve guide-configuring-plugins.apt

manaswinidas:
As suggested by @michael-o in #97


hboutemy:
thank you @manaswinidas and @elharo

PR #111, opened 2019-12-15
Remove java 6 instructions

elharo:
@hboutemy Given that we're now discussing whether we can require Java 8, I think it's safe to remove instructions on whether and how it's OK to require Java 6 in a plugin.


khmarbaise:
Currently we have only decided to use JDK8 for Maven Core 3.7.0+ not yet talked/discussed about plugins etc. to lift the minimum as well...

elharo:
Yes, but do we require special steps for those plugins to use Java 6? I think we're assuming at least java 7 everywhere.

khmarbaise:
Yes of course this is correct. So yes JDK6 is obsolete. No one should use JDK6 to develop a plugin ...anymore ...

slachiewicz:
Maybe we can only rename/replace Java 6 with Java 7 (plus redirect from old one)? If we assume 7 all information is still valid.

elharo:
Thanks. Looks like I'm not listed as a committer in this repo yet so someone else will have to merge.

PR #112, opened 2019-12-15
Remove subversion section

elharo:
@hboutemy as we're now using github

### DIFF
--- a/content/apt/developers/committer-environment.apt
+++ b/content/apt/developers/committer-environment.apt
@@ -1,5 +1,5 @@
  ------
- Developers centre - Commiter Environment
+ Developer centre - Committer Environment
  ------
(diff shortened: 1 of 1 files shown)


khmarbaise:
Hm...I'm not sure if it's good idea to remove SVN information cause at the moment the web sites are committed in SVN repos ...and if you do a release you need to have svn correctly running to commit the final state of the published sites https://maven.apache.org/developers/website/deploy-component-reference-documentation.html#Publishing_versioned_component_reference_documentation

elharo:
OK, we still need to do a update/cleanup pass on this page then. I don't think these instructions are fully current.

slachiewicz:
I think we use svn to publish versions of asf pages during release.

PR #116, opened 2019-12-24
[MNG-6851] change "property" to "system property"

kabike:
because project property can not activate profile


jira-importer:
Resolve #969

PR #121, opened 2019-12-31
Using an SPDX identifier as the license name is recommended.

mkarg:
Using SPDX identifiers as license names makes it much easier to unambiguously identify the used license, in particular by tools like the [Mojohaus License Maven Plugin](https://www.mojohaus.org/license-maven-plugin/) (e. g. because the configuration will be much shorter when only SPDX identifiers are used).


mkarg:
@khmarbaise Please consider this proposed best practice, which in fact is already applied by many people. :-)

PR #137, opened 2020-02-29
Add mojo conventions

chonton:
In the plugin developer's "Your first mojo" guide, add a section enumerating several best practices for mojos and provide an example.

### DIFF
--- a/content/apt/guides/plugin/guide-java-plugin-development.apt
+++ b/content/apt/guides/plugin/guide-java-plugin-development.apt
@@ -335,7 +335,7 @@ mvn archetype:generate \
 
 ** Mojo Conventions
 
-  Since plugins are often configured in a organizational or enterprise parent poms, there are
+  Since plugins are often configured in organizational or enterprise parent poms, there are
   conventions which make re-configuration in an inheriting project easier.
 
   [[1]] Provide a skip parameter for each goal.  This enables disabling a plugin in a child project
@@ -344,7 +344,7 @@ mvn archetype:generate \
   [[2]] Provide a default phase for each goal to eliminate the need to specify the phase in the <<<pom.xml>>>.
 
   [[3]] Some configuration names are quite common; e.g. 'includes', 'excludes', 'skip'.  To prevent
-  accidentally configuring multiple mojos sharing a common property name, use a your plugin's prefix
+  accidentally configuring multiple mojos sharing a common property name, use your plugin's prefix
   as a 'namespace' for your plugin's configuration names.
 
   []
--- a/content/apt/guides/plugin/guide-java-plugin-development.apt
+++ b/content/apt/guides/plugin/guide-java-plugin-development.apt
@@ -333,6 +333,50 @@ mvn archetype:generate \
   of the mojo parameter from the command line by referencing a system property that the user sets
   via the <<<-D>>> option.
 
+** Mojo Conventions
+
+  Since plugins are often configured in a organizational or enterprise parent poms, there are
+  conventions which make re-configuration in an inheriting project easier.
+
+  [[1]] Provide a skip parameter for each goal.  This enables disabling a plugin in a child project
+  with a simple configuration parameter or property.
+
+  [[2]] Provide a default phase for each goal to eliminate the need to specify the phase in the <<<pom.xml>>>.
+
+  [[3]] Some configuration names are quite common; e.g. 'includes', 'excludes', 'skip'.  To prevent
+  accidentally configuring multiple mojos sharing a common property name, use a your plugin's prefix
+  as a 'namespace' for your plugin's configuration names.
+
+  []
+
++-----+
+@Mojo( name = "sayhi", defaultPhase = LifecyclePhase.INITIALIZE )
+public class GreetingMojo extends AbstractMojo
+{
+    /**
+     * Skip execution of the sayhi goal
+     */
+    @Parameter( property = "sayhi.skip", defaultValue = "false" )
+    private boolean skip;
+
+    /**
+     * The greeting to display.
+     */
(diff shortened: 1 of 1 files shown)


content/apt/guides/plugin/guide-java-plugin-development.apt
```diff
@@ -333,6 +333,50 @@ mvn archetype:generate \
   of the mojo parameter from the command line by referencing a system property that the user sets
   via the <<<-D>>> option.
 
+** Mojo Conventions
+
+  Since plugins are often configured in a organizational or enterprise parent poms, there are
```
elharo: delete "a" before organizational

content/apt/guides/plugin/guide-java-plugin-development.apt
```diff
@@ -333,6 +333,50 @@ mvn archetype:generate \
...
+  conventions which make re-configuration in an inheriting project easier.
+
+  [[1]] Provide a skip parameter for each goal.  This enables disabling a plugin in a child project
+  with a simple configuration parameter or property.
+
+  [[2]] Provide a default phase for each goal to eliminate the need to specify the phase in the <<<pom.xml>>>.
+
+  [[3]] Some configuration names are quite common; e.g. 'includes', 'excludes', 'skip'.  To prevent
+  accidentally configuring multiple mojos sharing a common property name, use a your plugin's prefix
```
elharo: a your --> your

rfscholte: I don't like these proposed conventions.
1. a skip parameter is actually a bad sign. It might expose a problem with your project plugins. And sometimes it doesn't make sense, especially goals that are executed directly from commandline (why does `dependency:tree` have a skip parameter). 
2. there's not always default phase, you should check it per goal it it makese sense.

chonton: 1. The plugins I am most familiar with have a skip parameter.

Examples from apache maven:
- compiler: maven.main.skip, maven.test.skip
- deploy: maven.deploy.skip
- enforcer: enforcer.skip
- failsafe: maven.test.skip, skipTests, skipITs
- install: maven.install.skip
- invoker: invoker.skip
- jar: maven.test.skip
- pmd: pmd.skip
- resources: maven.resources.skip, maven.test.skip
- site: maven.site.skip
- surefire: maven.test.skip, skipTests

Examples from elsewhere:
- docker-maven-plugin (fabric8): docker.skip, docker.skip.build, docker.skip.push, docker.skip.run, docker.skip.tag, docker.skip.machine
- duplicate-finder-maven-plugin: duplicate-finder.skip
- exists-maven-plugin: exists.skip
- fmt-maven-plugin (coveo): fmt.skip
- jacoco: jacoco.skip 
- japicmp: japicmp.skip
- sortpom-maven-plugin: sort.skip
- spotbugs-maven-plugin: spotbugs.skip

This convention already exists.  This convention has emerged over years from developers' needs.  I believe that we should document a pattern that has been found by experience.


2. I agree that default phase do not always make sense. I will modify the wording.

rfscholte: The fact that there is a skip parameter doesn't mean that it is a best practice. It is more a sign that something else is broken.
People are even asking to add a skip parameter to somthing like maven-shade-plugin, but that has impact on the generated artifact.  And explain why the maven-jar-plugin has a skip parameter, the whole purpose is to create a jar, right?
I'm planning to remove several skip parameters mentioned, the maven-compiler-plugin being the first.

chonton: What do you believe is the "something else" that is broken?  Unless the "something else" can be fixed, developers will continue to want 'skip' parameters.  Two use cases for skip are:
1. Skip a goal that takes a long time by providing a system property on command line.  I need to quickly and iteratively execute my pipeline. (compile or skipITs are examples of this use case)
2. Specialization: skip a goal that was inherited from parent that doesn't make sense for this module.

rfscholte: 1. I haven't seen compile being slow, so I don't see the need for skip. Integration tests I agree, as it doesn't influence the artifact. Either skip or a dedicated profile is fine.
2. this is true and and should be fixed in Maven. Now inheritence is set to true/false, but it should be more finegrained. (can't find the matching ticket to improve this)

We could go over the complete list of existing skip parameters, but in the end it is a quick workaround and should not be seen as best practice.

chonton: [MNG-3102](https://issues.apache.org/jira/browse/MNG-3102) identified the pattern and asked for a general solution in 2008.  [MNG-3321](https://issues.apache.org/jira/browse/MNG-3321) proposed a general solution for maven core in 2011.  These patches were never merged.  This lack of activity plus the desire to support older maven core versions for an extended adoption period suggests to me that skip parameters will be with us for decades.

rfscholte: If you don't want a plugin's goal to be executed, don't make it part of the buildplan at all. A skip parameter would still initalize the whole plugin, get its required dependencies, etc. and after that the execution starts with: `if(skip) return;` This is far  from efficient.
If you don't want a plugin's goal to be executed, use profiles.

Hence, I don't want this mentioned as best practices on the Apache Maven pages. If you write your own plugin, and somebody needs to ability to skip it in a lazy way, let him provide a PR on that plugin. It should not be there by default.

chonton:
any further changes requested?

elharo:
No particular changes. I think you need to get consensus on this from the Maven maintainers before a PR like this will be accepted.

PR #159, opened 2020-04-18
docs: minor grammar issues

elharo:
@michael-o


slachiewicz:
https://github.com/khmarbaise/maven-it-extension here we have more about testing

PR #164, opened 2020-05-10
[MNGSITE-407] remove old page

elharo:
@khmarbaise


slachiewicz:
For our site changes - no need to create jira every time

PR #165, opened 2020-05-11
docs: git migration is complete

elharo:
@olamy


slachiewicz:
Yes, we have only few remaining - but this info is longer interesting to wider forum.

PR #170, opened 2020-05-23
[MNGSITE-393] we do not support Maven 2

elharo:
@slachiewicz


elharo:
This doesn't change that. It just removes mention of Maven 2 from this page as part of a long effort to purge all Maven 2 from or main docs.

PR #191, opened 2020-08-04
docs: system dependencies are not always available

elharo:
@michael-o


content/apt/guides/introduction/introduction-to-dependency-mechanism.apt
```diff
@@ -172,9 +172,10 @@ Introduction to the Dependency Mechanism
    unit tests (src/test/java) but not in the model code (src/main/java).
 
  * <<system>>\
-   This scope is similar to <<<provided>>> except that you have to provide the JAR
-   which contains it explicitly. The artifact is always available and is not
-   looked up in a repository.
+   Maven does not add system-scoped dependencies to any classpath.
```
michael-o, line 175: If it is not added to a classpath, what's the point in having them?

elharo: They're usually (though not always) in the classpath but Maven doesn't put them there. The JVM (or something else) does.

michael-o: But at least for compilation purposes it is in the classpath or maybe at runtime available.

elharo: The keyword is "maybe". There's no guarantee it is, and the current text makes it sound like this is guaranteed. The case that made me pay attention to this was Guava doing a complicated dance to load src.zip for use in javaDoc:

```
   <profile>
      <id>srczip-parent</id>
      <activation>
        <file>
          <exists>${java.home}/../src.zip</exists>
        </file>
      </activation>
      <dependencies>
        <dependency>
          <groupId>jdk</groupId>
          <artifactId>srczip</artifactId>
          <version>999</version>
          <scope>system</scope>
          <systemPath>${java.home}/../src.zip</systemPath>
          <optional>true</optional>
        </dependency>
      </dependencies>
    </profile>
```

michael-o: I don't see the maybe in the modified text.

elharo: I was referring to the "maybe"in your comment. :-)

content/apt/guides/introduction/introduction-to-dependency-mechanism.apt
```diff
@@ -174,9 +174,10 @@ Introduction to the Dependency Mechanism
...
```
michael-o, line 177: I fo not agree with this. @rfscholte

michael-o (changes requested):

elharo:
Ping @rfscholte 

I still think this doc change is needed. The current documentation is simply incorrect.

content/apt/guides/introduction/introduction-to-dependency-mechanism.apt
michael-o: It is unfortunately that easy and it is not just the JDK

* https://issues.apache.org/jira/browse/MNG-1867
* https://issues.apache.org/jira/browse/MNG-6523
* The idiot's from SAP have hardoded the JCo JAR name and it won't start if you rename it
* JAR which people aren't allowed for reasons to upload to a local repo instance
* likely others

elharo: It's been a while but this came up a couple of project ago in my day job, where we spent a lot of effort untangling different kinds of dependency scopes and analyzing dependency graphs. The existing text is wrong in one key respect:

"Dependencies with the scope system are always available and are not looked up in repository." 

However, dependencies with the scope system are *not* "always available". Declaring a dependency such as 

```
<dependency>
          <groupId>org.apache.maven.plugins</groupId>
          <artifactId>maven-site-plugin</artifactId>
          <version>4.0.0-M7</version>
          <scope>system</scope>
</dependency>
```

doesn't magically make it available. This is likely just sloppy writing, and the intent was to document the correct behavior but that's not how the text reads  currently, especially to a dev not steeped in Maven dependency management. 

Let me take another stab at rephrasing this to take into account the points you bring up about non-classpath based resolution of dependencies.

elharo (commented):
PTAL

rfscholte:
LGTM

content/apt/guides/introduction/introduction-to-dependency-mechanism.apt
```diff
@@ -826,12 +826,14 @@ Introduction to the Dependency Mechanism
...
- JDK, but were available as separate downloads earlier. Typical examples are
- the JDBC standard extensions or the Java Authentication and Authorization
+ Dependencies with the scope <system> are not looked up in the Maven repository system.
+ Instead the `dependency` element contains a `systemPath` pointing to a jar on the local
+ file system.
+ 
+ The system scope is commonly used to tell Maven about dependencies provided by the JDK or the VM. 
+ System dependencies are especially useful for resolving dependencies on artifacts which
+ are now provided by the JDK, but were available as separate downloads earlier. Typical
+ examples are the JDBC standard extensions or the Java Authentication and Authorization
```
michael-o, line 836: I don't even know whether the cae with JDBC and JAAS is still true. I'd try to avoid giving example which are invalid in a couple of years.

michael-o (changes requested):

pom.xml
```diff
@@ -55,7 +55,8 @@
 
   <properties>
     <maven.site.path>maven-site</maven.site.path>
-    <project.build.outputTimestamp>x</project.build.outputTimestamp><!-- disable Reproducible Builds that makes the site build time fixed at last release time -->
```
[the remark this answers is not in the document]
elharo: Yes, but necessary or the build breaks. I am less find of how spotless has been integrated every day. :-(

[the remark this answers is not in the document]
michael-o: Spotless 1🤣

content/apt/guides/introduction/introduction-to-dependency-mechanism.apt
elharo: I think it is. Per Oracle docs, "The Java Authentication and Authorization Service (JAAS) was introduced as an optional package (extension) to the Java 2 SDK, Standard Edition (J2SDK), v 1.3. JAAS was integrated into the J2SDK 1.4." Unless Oracle decides to remove it again, this statement seems likely to be true for the indefinite future. 

Though now that I think about it, I don't really see how this helps to to tell Maven about dependencies provided by the JDK or the VM. If you're in a VM that includes them, this is unnecessary and if you're not then a system dependency doesn't help since it's not there. Maybe if you're compiling with JDK 1.5 but targeting 1.3 or something like that?

michael-o: Even if, i don't know wether this really helps.

elharo: I'm OK with removing this paragraph completely. The case that makes me pause is tools.jar. If I recall correctly that is not in the VM and also is not available from the Maven repo system. Is there any way to load that that is not a system dependency?

michael-o: I think if you custom-bundle Maven with your deps, e.g., extensions or so.

elharo: I'm not quite sure what you mean. Do you mean some sort of fat jar? Or perhaps a custom version of Maven? Either way this seems much more complex than a system dependency that points to tools.jar. 

To move this PR forward, what specific change in the current text do you propose?

michael-o: My example was just have a custom JAR which you distribute across your company. That's it.

My proposal is to simply remove all examples regarding `system` and provide an abstract description/definition of this feature.

slachiewicz: Agree, let's remove references to jdbc. Nowadays Oracle jdbc is normally available

elharo: done

slachiewicz (approved):

PR #198, opened 2020-09-10
Update index.xml.vm

zzahti:


mthmulders:
This merge request makes no sense.

elharo:
That's kind. It's an obvious and poorly done spam attempt. Anyone feel like blocking this user? Or can we report them to Github?

PR #211, opened 2020-10-22
Fixes custom header configuration example

mattjohnson:
Fixes headers configuration to use property instead of header


hboutemy:
issue/fix found with https://github.com/mattjohnson/maven_custom_headers Maven setup and checking headers at server level

tested with recent Maven version, no idea if there was a parameter renaming during Maven code evolution, or if the doc was just not updated while implementing what looked a proposal https://github.com/apache/maven-site/commit/8e6560a54941610a688f220d641ea760dca20a60

olamy (approved):

slachiewicz (approved):

PR #228, opened 2021-02-12
[MNGSITE-430] add dependency POMs

hboutemy:


aheritier (approved):

rfscholte:
I suggest to split this up into two: 
- The repository resolution order for a local/reactor pom
- The repository resolution order for a dependency pom

It must be very clear that transitive dependencies are never effected by repos of their calling pom.

hboutemy:
@rfscholte local POM does not have any resolution: it is locally on disk, from Git
resolution is always for dependencies

> transitive dependencies are never effected by repos of their calling pom.
they are affected by repos of their calling pom, that is exactly what resolution order is about

rfscholte:
The last description is still not clear for me. I think what we're trying to say: repositories within context, i.e the reposities of the pom that defined this pom as a dependency.

hboutemy:
yes, the last description is the hard one where we need creativity...

"POMs from dependency path to the artifact"?

rfscholte:
Sometimes images say more than words. Looks like this one should have a picture of a graph to support the logic.

hboutemy:
yes, we can add such an image in a future step

jira-importer:
Resolve #997

PR #236, opened 2021-04-30
[MNGSITE-444] Add missing reactor documentation

MartinKanters:
This PR describes most of the changes to the reactor that are coming in Maven 4.
@mthmulders and I also extended the existing docs.

@elharo We would like to ask you for a review at least, as you are a native speaker.


michael-o:
I would prefer no priorpritary format/application to be necessary to draw stuff. SVG is preferred or as @hboutemy did a dep over view with LO Draw.

mthmulders:
> I would prefer no priorpritary format/application to be necessary to draw stuff. SVG is preferred or as @hboutemy did a dep over view with LO Draw.

I found some .ai files (think it's Adobe Illustrator) so I thought it would be no problem. Replaced it with ODP which I believe is an open format.

elharo (commented):
Minor style nits as requested. I'm not familiar enough with this work to judge whether the text is accurate.

content/apt/guides/mini/guide-multiple-modules.apt
```diff
@@ -37,11 +39,25 @@ Guide to Working with Multiple Modules
...
 
-    * Builds the selected projects in order
+    * Selects which modules to build
 
-** Reactor Sorting
+    * Builds the selected modules in order
+
+** Collecting Modules
+
+  Module collection starts from one aggregate project. That project defines the modules of which it consists using the <<<\<modules\>>>> element.
```
hboutemy, line 50: "aggregate project" or "aggregating project"?

content/apt/guides/mini/guide-multiple-modules.apt
```diff
@@ -59,25 +75,117 @@ Guide to Working with Multiple Modules
 
   Note that only "instantiated" references are used - <<<dependencyManagement>>> and <<<pluginManagement>>> elements do not cause a change to the reactor sort order.
 
-** Command Line Options
+** Selecting Modules
```
hboutemy, line 78: perhaps add a "Maven 4" tip here, because IIUC, this is the new part

content/apt/guides/mini/guide-multiple-modules.apt
```diff
@@ -59,25 +75,117 @@ Guide to Working with Multiple Modules
...
+  This prefix should always go after the other prefixes.
 
-    * <<<--resume-from>>> - resumes a reactor from the specified project (e.g. when it fails in the middle)
+*** Relationships between modules
 
-    * <<<--also-make>>> - build the specified projects, and any of their dependencies in the reactor
+  Modules inside a project can have two types of relationships: parent/child and dependency/dependent.
 
-    * <<<--also-make-dependents>>> - build the specified projects, and any that depend on them
+  When selecting a parent (aggregate), Maven automatically selects the child modules as well. Similarly, Maven excludes child modules of an excluded parent (aggregate).
```
hboutemy, line 106: aggregate -> aggregator

MartinKanters:
@elharo @hboutemy Thanks for your feedback! I've processed all points.

content/apt/guides/mini/guide-multiple-modules.apt
MartinKanters: Well yeah, that's a hard one, there are definitely a couple of things changed in this section, but not everything is new. For example, including/excluding projects with `-pl` was always possible, but optional project (`?` prefix) selection was added. 
We tried to come up with some somewhat more 'advanced' changes between 3 and 4 in the table at the bottom of the page. I think it will become too chaotic to link every new thing to a "Maven 4 tip". Some advanced (breaking) changes are listed in the table below.

We have briefly thought of keeping the [current guide](https://maven.apache.org/guides/mini/guide-multiple-modules.html) on a separate page. And linking to and from that from this new page... a.k.a. "These docs contain Maven 4 features, looking for Maven 3 docs? -> <insert link>" and vice versa. Do you think that would be valuable?

MartinKanters: Discussed offline that it might be a better idea to keep the current (Maven 3) guide intact and move this PR content towards the docs in [maven-embedder](https://github.com/apache/maven/tree/master/maven-embedder/src/site/apt#L0-L0). We can then point from the Maven 3 docs to the details (which are hosted in maven-embedder).

mthmulders: We gave it a second thought, and we're not so sure about moving it to Maven Embedder or Maven Core. In its current form, the page covers everything a Maven user might need: the concepts and how they work together as well as how to leverage them by passing the CLI flags.

If we'd move the page to Embedder or Core, it would be strange to have the CLI flags mentioned there, as CLI flags only exist in the maven-cli module. It would effectively split the page in two: "here's how it works" vs "here's how you can use it". We feel it would make it _harder_ for our users to find the relevant documentation.

So, we propose to keep it in one page, targeting Maven 4, which would live next to the existing page that targets Maven 3. The two pages will cross-reference each other:

image

And the two pages will both show up in the menu:

image

PR #249, opened 2021-07-27
Minor improvement to English fluency

cpsauer:
Fixed some unusual phrasing I saw while reading the docs. 
Goal is to leave things even better than I found them.


michael-o (approved):

cpsauer:
Wow. That was fast. Awesome. Thanks!

PR #272, opened 2021-11-06
Clarify HEAD being the default for scm.tag

kwin:


hboutemy:
you're right https://maven.apache.org/ref/3.8.3/maven-model/maven.html#class_scm

thank you

PR #278, opened 2021-12-09
Fix url syntax for javaalmanac.io

slawekjaranowski:


hboutemy:
stupid me: thank you :)

PR #294, opened 2022-03-07
typo in "direct plugin goal invocation" example

bhamail:
The example given for invoking a direct plugin goal execution has a typo. I think this fix has the correct syntax.


michael-o (approved):

slawekjaranowski:
@bhamail thanks

PR #311, opened 2022-07-18
Update guide-attached-tests.apt

fkleedorfer:
The guide is incomplete.  Surefire does not pick up the tess in the attached jar on its own, it has to be configured to to so, and its provider autodetection does not work with attached jars, so it has to be nudged.

With this change, the guide would have saved me an hour.

### DIFF
--- a/content/apt/guides/mini/guide-attached-tests.apt
+++ b/content/apt/guides/mini/guide-attached-tests.apt
@@ -54,7 +54,7 @@ Guide to using attached tests
   </build>
 </project>
 
-The attached test JAR can be installed and deployed like any other Maven artifact with 
+The attached test JAR can be installed and deployed like any other Maven artifact with
 
 To use the attached test JAR, specify a dependency on the main
 artifact with a specified type of <<<test-jar>>> and the <<<classifier>>>.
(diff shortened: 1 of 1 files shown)


content/apt/guides/mini/guide-attached-tests.apt
```diff
@@ -90,14 +93,42 @@ mvn deploy
...
       <version>1.0-SNAPSHOT</version>
       <classifier>tests</classifier>
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    
+    <dependency>
+      <!-- 
+        only needed if the test provider is not correctly autodetected by surefire 
+        ... and you are using junit-jupiter (othewise, replace with your test framework)
```
elharo, line 106: Otherwise

fkleedorfer: ok

content/apt/guides/mini/guide-attached-tests.apt
```diff
@@ -81,7 +81,10 @@ mvn deploy
 * Using the attached test JAR
 
  In order to use the attached test JAR that was created above you simply specify a dependency on the main
- artifact with a specified type of <<<test-jar>>> and the <<<classifier>>>.
+ artifact with a specified type of <<<test-jar>>> and the <<<classifier>>>, and tell the surefire plugin to scan 
+ that dependency for tests. Also, you might have to tell surefire which provider you are using as it might not be 
+ able to autodetect it from the <<<dependenciesToScan>>>. In that case, simply add a test-scoped dependency to
```
elharo, line 86: "as it might not be 
 able to " --> "if it cannot"

elharo (commented):
If you can update the diff with the latest from head, we should be able to merge this. Thanks.

fkleedorfer:
done

slawekjaranowski:
Where is link to this site, from Maven site?

IMHO should be content of surefire plugin documentation not Maven itself.

fkleedorfer:
If that is indeed the way to go, it should be another PR then, no?

slawekjaranowski:
by the way did you check output of site? 
https://maven.apache.org/guides/mini/guide-attached-tests.html

PR #325, opened 2022-10-07
Remove property from parameter for Map and Properties

slawekjaranowski:
There is no converters from String to Map, Properties So we can not use property for parameter of this type

Tested with Maven 3.8.6 and I have:

```
Cannot assign configuration entry 'properties' with value '${myProp}' of type java.lang.String to property of type java.util.Properties 
```

```
 Cannot assign configuration entry 'myMap' with value '${myMap}' of type java.lang.String to property of type java.util.Map
```

### DIFF
--- a/content/apt/guides/mini/guide-configuring-plugins.apt
+++ b/content/apt/guides/mini/guide-configuring-plugins.apt
@@ -345,7 +345,7 @@ public class MyAnimalMojo
 
 +-----+
 ...
-    @Parameter(property = "myMap")
+    @Parameter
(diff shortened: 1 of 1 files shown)


michael-o (approved):

kwin:
Maybe an explicit hint would be good.

kwin (approved):

slawekjaranowski:
> Maybe an explicit hint would be good.

Should be an another paragraph about using property with `Parameter`.

PR #326, opened 2022-10-10
[MNGSITE-496] Outline all means to load extensions and clarify

kwin:
differences

### DIFF
--- a/content/apt/examples/maven-3-lifecycle-extensions.apt.vm
+++ b/content/apt/examples/maven-3-lifecycle-extensions.apt.vm
@@ -28,7 +28,7 @@
 
 Example: Using Maven 3 lifecycle extension
 
-* Lifecyle Participation
+* Lifecycle Extension Points
 
   You can extend multiple classes depending on your needs:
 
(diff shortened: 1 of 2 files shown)


cstamas (approved):
But POM has extensions element as well https://maven.apache.org/pom.html#Extensions

cstamas:
Maybe state clearly that [afterSessionStart](https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/AbstractMavenLifecycleParticipant.java#L63) does NOT work in "late" mode for obvious reasons.

kwin:
> But POM has extensions element as well https://maven.apache.org/pom.html#Extensions

I think I mentioned those at https://github.com/apache/maven-site/pull/326/files#diff-bcbf6d6b4afe0d44d1c378f922812615b44429cb439b08fcb503e47446c138f4R83.

kwin:
> Maybe state clearly that [afterSessionStart](https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/AbstractMavenLifecycleParticipant.java#L63) does NOT work in "late" mode for obvious reasons.

Is that not clear enough: https://github.com/apache/maven-site/pull/326/files#diff-d57f2ffce98d7dc59a0359835112b62a6bb7213f279eae4c60f67a6a9ef5638fR125? For me that is an aspect for extension writers and not so much for extension consumers?

cstamas:
oh, i missed those, sorry (I hate archaic APT format). Yes, is important for ext authors, not much for consumers.

jira-importer:
Resolve #1055

PR #330, opened 2022-10-14
[MNG-7566] Plugin descriptor 1.1.0 adding element "requiredJavaVersion"

kwin:


kwin:
@hboutemy I found https://issues.apache.org/jira/browse/MNG-5624 but I don't really see why the XSD should not be published. Can you share why you think XSD should not be published?
IMHO also the java code should be generated with Modello. I am gonna give it another try with the new plugin descriptor 1.1.

kwin:
I meanwhile figured out why the XSD is wrong and created https://github.com/codehaus-plexus/modello/issues/264 for that.

slachiewicz:
@kwin is this still valid?

kwin:
v1.1 was never formalized and has been meanwhile overtaken by v2.

PR #350, opened 2022-12-06
Add link to shared-resources with code style formatters

slawekjaranowski:


content/apt/developers/conventions/code.apt
```diff
@@ -30,9 +30,13 @@ Maven Code Style And Code Conventions
 
  This document describes the rules for how the sources should be formatted
  in order to improve consistency, readability and maintainability.
+
  As the formatting is automatically enforced or even applied with {{{https://github.com/diffplug/spotless/tree/main/plugin-maven}spotless-maven-plugin}} for all projects
  using {{{/pom/maven/index.html}Maven Project Parent POM 38 or newer}} developers usually don't need to care and the following sections are just for informational purposes.
 
+ For comfortable work, you can import the code style formatter for your IDE from
```
kwin, line 37: This requires some more explanation, as there are two files for IntelliJ but only one for Eclipse....

slawekjaranowski: maybe README in shared repo will be better ...

we have two for IJ - for old and for new format
eclipse has only old .... new formatter should be prepared

cstamas (approved):

PR #353, opened 2022-12-09
[INFRA] add autolink refs

bmarwell:


.asf.yaml
```diff
@@ -9,6 +9,28 @@ github:
...
+    - MPLUGIN
+    - MWRAPPER
+    - MSHARED
+    - MSKINS
+    - MPOM
+    - ARCHETYPE
+    - ARCHETYPES
+    - INFRA
+    - MASSEMBLY
+    - FELIX
```
kwin, line 32: Shouldn’t we restrict the list to Maven JIRA projects? Do you really expect a FELIX JIRA ID appear in the commit messages of the Maven Site?

bmarwell: I only added projects I found in the log. Just do sth like
git log | grep -iE -e "[A-Z]+-[0-9]+" and you will find all of the listed projects (for whatever reason).
Granted, most of them weren't mentioned since 2015...

slawekjaranowski (approved):

PR #371, opened 2023-01-19
Plugin testing: remove verifier example and instead link to verifier Getting Started guide

psiroky:
* addCliOption() and resetStreams() are deprecated in the current verifier version (2.0.0-M1)

### DIFF
--- a/content/apt/plugin-developers/plugin-testing.apt
+++ b/content/apt/plugin-developers/plugin-testing.apt
@@ -114,72 +114,9 @@ Integration/Functional testing
 
 * maven-verifier
 
- maven-verifier tests are run using JUnit or TestNG, and provide a simple class allowing you to launch Maven and assert on its log file and built artifacts.  It also provides a ResourceExtractor, which extracts a Maven project from your src/test/resources directory into a temporary working directory where you can do tricky stuff with it.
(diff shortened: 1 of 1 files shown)


content/apt/plugin-developers/plugin-testing.apt
```diff
@@ -144,16 +144,16 @@ public class TrivialMavenVerifierTest extends TestCase
...
+         * as strings, similarly as we would when calling Maven
+         * directly from command line. This is handy for things like
          * redefining the local repository if needed. In
          * this case, we use the -N flag so that Maven won't
          * recurse. We are only installing the parent pom to
          * the local repo here.
          */
-        List cliOptions = new ArrayList();
-        cliOptions.add( "-N" );
-        verifier.executeGoal( "install" );
```
psiroky, line 156: Unless I am missing something, the original code did not make much sense, because the `cliOptions` list was never used by verifier. With the newer verifier version, there is no need for a list since the method uses variable length arg).

slawekjaranowski:
I would go a step forward, general page should only contains general description about tools, and links to pages maintained by tools. 
In Verifier case we have: https://maven.apache.org/shared/maven-verifier/getting-started.html

When we change somethings in verifier I afraid that we forgot to update this site again.

psiroky:
@slawekjaranowski I removed the example and just linked the Getting Started guide. Please take another look if this looks like something you had in mind. Thanks!

PR #372, opened 2023-01-21
Fix code blocks that wrap until bottom of page.

agreedSkiing:
This fixes markdown formatting on page so that table of content linking works again.

Edit done through GitHub editor.

P.S. Markdown code blocks should end with ` ``` `. If a block ends with ` ```xml ` the parser assumes that a new code block should start.


slawekjaranowski:
@agreedSkiing Thanks, looks better

PR #378, opened 2023-02-06
Update download page

cstamas:
Now both, latest stable Maven 3.x and Maven 4.x require Java 8, make it written so. Also, current installation size for 3.9.0 is 11MB :)


cstamas:
@jorsol was faster https://github.com/apache/maven-site/pull/377

PR #382, opened 2023-02-09
[MNGSITE-507] Converted .apt documents to Markdown

slachiewicz:


kwin:
https://maven.apache.org/pom.html#packaging looks weird, pretty sure "packaging" hasn't been a headline before.

kwin:
Also https://maven.apache.org/guides/mini/guide-configuring-plugins.html has incorrect escape characters

For such a big change I would really appreciate to involve some more devs in the PR to look at that before merging.

kwin:
TOC broken in https://maven.apache.org/guides/mini/guide-configuring-plugins.html. On that page some headlines are not rendered as such ("Mapping Maps") and numbered lists are broken (https://issues.apache.org/jira/browse/DOXIA-693). Probably that affects a lot more pages.

jira-importer:
Resolve #869

PR #405, opened 2023-03-30
Remove m-pdf-plugin - no longer works with m-site-p 4.x/Doxia 2.x

slachiewicz:


pom.xml
```diff
@@ -426,11 +421,11 @@
   </reporting>
 
   <profiles>
+    <!--
```
kwin, line 424: Please remove instead of commenting out

slachiewicz: fixed

PR #406, opened 2023-03-30
Update introduction-to-the-pom.apt

gusega:
update merged elements


content/apt/guides/introduction/introduction-to-the-pom.apt
```diff
@@ -137,7 +137,9 @@ Introduction to the POM
 
  Elements in the POM that are merged are the following:
 
- * dependencies
+ * dependencies and dependencyManagement
+
+ * variables
```
slawekjaranowski, line 142: There is no such sections in pom like `variables` - do you mean `properties`?

PR #412, opened 2023-04-23
spotless fix

elharo:
somebody must have updated without running `mvn test`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,8 @@
 
   <properties>
     <maven.site.path>maven-site</maven.site.path>
-    <project.build.outputTimestamp>x</project.build.outputTimestamp><!-- disable Reproducible Builds that makes the site build time fixed at last release time -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -55,7 +55,8 @@
 
   <properties>
     <maven.site.path>maven-site</maven.site.path>
-    <project.build.outputTimestamp>x</project.build.outputTimestamp><!-- disable Reproducible Builds that makes the site build time fixed at last release time -->
+    <project.build.outputTimestamp>x</project.build.outputTimestamp>
+    <!-- disable Reproducible Builds that makes the site build time fixed at last release time -->
```
slawekjaranowski, line 59: should be one line upper

elharo: I'm not sure what you mean. In any case, this is exactly how spotless changed the existing file, nothing else.

PR #435, opened 2023-07-04
Fix broken link to Gitea's repository manager support

wetneb:
I have not checked the other links.


slawekjaranowski:
@wetneb thanks

PR #441, opened 2023-07-23
[MNG-7850] Add example for low-level resolver configuration

slawekjaranowski:


content/markdown/guides/mini/guide-resolver-transport.md
```diff
@@ -99,6 +99,60 @@ for example to allow the user to tell Maven how long to wait before giving up on
   </servers>
 </settings>
 ```
+## Low-level Resolver configuration
+
+All configuration items mentioned in  [Configuration Options](https://maven.apache.org/resolver/configuration.html) 
+which is not mapped by server/configuration tag must be provided in special way.
```
cdprete, line 105: What does `which is not mapped by server/configuration tag` exactly mean?

content/markdown/guides/mini/guide-resolver-transport.md
```diff
@@ -99,6 +99,60 @@ for example to allow the user to tell Maven how long to wait before giving up on
...
+  <profiles>
+    <profile>
+      <id>resolver-config</id>
+      <properties>
+        <aether.dependencyCollector.impl>bf</aether.dependencyCollector.impl>
+        <!-- set preemptiveAuth for all servers -->
+        <aether.connector.http.preemptiveAuth>true</aether.connector.http.preemptiveAuth>
+        
+        <!-- set preemptiveAuth for only for server with id - serverId -->
+        <aether.connector.http.preemptiveAuth.serverId>true</aether.connector.http.preemptiveAuth.serverId>
```
cdprete, line 122: That `serverId` makes it look like it's another property. 
I know it's not the case, but it's very confusing.

cdprete (changes requested):

content/markdown/guides/mini/guide-resolver-transport.md
slawekjaranowski: Like items mentioned above in this document 
- Custom HTTP Headers - `configuration/httpHeaders`
- Timeouts  `configuration/connectTimeout` ...

content/markdown/guides/mini/guide-resolver-transport.md
slawekjaranowski: my mistake - should be a `repositoryId`
some of properties can be applied globally or by repository, like in example
`aether.connector.http.preemptiveAuth` - is applied for all repositories
and the same property `aether.connector.http.preemptiveAuth.repoisitoryId` - is applied only for specific repository

cdprete: I would then suggest another format, for the documentation only, for such properties like `aether.connector.http.preemptiveAuth[.<repositoryId>]` and explain that the `<repositoryId>` must actually be the `<server><id>...</id></server>` defined in the settings.
As you can see, having different names for the same thing makes things complicated.
I would actually stick to `serverId` as name instead of `repositoryId` to avoid this confusion (and 'cause it's probably cheaper than changing the `<server>` block which would be a breaking change).

cstamas: Many resolver configuration supports this "per remote repo" config, see https://maven.apache.org/resolver/configuration.html and table last column "Supports Repo ID Suffix"

cstamas: Agree about "alt format", as current is simply misleading, looks really like some new property. I am fine with proposed `aether.connector.http.preemptiveAuth[.<repositoryId>]` and also maybe `aether.connector.http.preemptiveAuth.$REPOSITORY_ID`? etc

cdprete: Hi @cstamas.
Yes, I saw that.
I was just suggesting to change the naming convention to be more consistent, ;)

slawekjaranowski:
ok, next try - @cdprete @cstamas please look at last version.

cstamas (approved):

slawekjaranowski:
By the way it is some of workaround for me ... we should extend model for settings.xml to cover new configuration items, like proposition in: 
https://issues.apache.org/jira/browse/MNG-7849

low-level items should be used only in special case - no by standard configuration.

content/markdown/guides/mini/guide-resolver-transport.md
cdprete: > Agree about "alt format", as current is simply misleading, looks really like some new property. I am fine with proposed `aether.connector.http.preemptiveAuth[.<repositoryId>]` and also maybe `aether.connector.http.preemptiveAuth.$REPOSITORY_ID`? etc

Seems like we replied at the same moment and I didn't see this comment.
Both the formats are fine for me personally, as soon as end users know that `repositoryId` = `$REPOSITORY_ID` = `<server><id>...</id>/<server>`.

Maybe `$REPOSITORY_ID` may be misleading since people may think you necessarily need an environment variable for it (although then it would be `${env-REPOSITORY_ID}`).

cdprete:
> By the way it is some of workaround for me ... we should extend model for settings.xml to cover new configuration items, like proposition in: https://issues.apache.org/jira/browse/MNG-7849
> 
> low-level items should be used only in special case - no by standard configuration.

Agree.
To be honest, I personally prefer the old way on how these things were set up.
In the old way, it was clear without a doubt that for this `<server>` block I'm applying this `<configuration>` block.

content/markdown/guides/mini/guide-resolver-transport.md
cdprete: Ok, I would then rewrite the sentence as:
> All configuration items mentioned in  [Configuration Options](https://maven.apache.org/resolver/configuration.html) 
which **are** not mapped by server/configuration tag (e.g.: like for [Custom Headers](https://maven.apache.org/guides/mini/guide-resolver-transport.html#custom-http-headers) and [Connection Timeouts](https://maven.apache.org/guides/mini/guide-resolver-transport.html#timeouts)) must be provided in special way.

PR #445, opened 2023-09-04
Update pom.apt.vm

oliveryasuna:
Replaced link to "Maven 2 Pom Demystified".

The current link no longer works. I replaced it with the latest working Internet Archive link.

### DIFF
--- a/content/apt/pom.apt.vm
+++ b/content/apt/pom.apt.vm
@@ -2033,4 +2033,4 @@ mvn help:active-profiles
 ======================
 
   Aspects of this guide were originally published in the
-  {{{http://www.javaworld.com/javaworld/jw-05-2006/jw-0529-maven.html}Maven 2 Pom Demystified}}.
+  {{{https://web.archive.org/web/20100327140148/https://www.javaworld.com/javaworld/jw-05-2006/jw-0529-maven.html}Maven 2 Pom Demystified}}.


content/apt/pom.apt.vm
```diff
@@ -2033,4 +2033,4 @@ mvn help:active-profiles
 ======================
 
   Aspects of this guide were originally published in the
-  {{{http://www.javaworld.com/javaworld/jw-05-2006/jw-0529-maven.html}Maven 2 Pom Demystified}}.
+  {{{https://web.archive.org/web/20100327140148/https://www.javaworld.com/javaworld/jw-05-2006/jw-0529-maven.html}Maven 2 Pom Demystified}}.
```
elharo, line 2036: Thanks for the catch. We don't necessarily have to go full Turabian here, but if the URL doesn't work we could use a more classical citation "The Maven 2 POM demystified
The evolution of a project model", Eric Redmond, JavaWorld.com, 05/29/2006

or something like that

oliveryasuna: Whatever you think makes sense. I'll let the usual maintainers take it from here.

cstamas: @elharo I'm with your solution, no need a link to archive to Maven2...

Bukama: As I updated many lines of the `pom.apt.vm`in my PR #601, I also changed that line to 

>   Aspects of this guide were originally published in the article "Maven 2 Pom Demystified The evolution of a project model" by Eric Redmond (JavaWorld.com, 2006-05-29).

I think it's okay for you, @oliveryasuna, as you have the desicion handed over to the maintainers (@elharo and @cstamas here) and I picked up their intention.

elharo: Do you need to push this change? I don't see it.

Bukama: > Do you need to push this change? I don't see it.

I integrated it in #601

elharo: so we should close this PR?

Bukama: > so we should close this PR?

Yes I think this PR (445) can be closed.

PR #455, opened 2023-09-24
remove vestigial ant files

elharo:
As best I can tell these are completely unused and unreferenced. Nothing breaks when I remove them. I tend to err on the side of tearing down Chesterton's fence. Worst case, if we l;ater discover a need for these this is easy to revert.


cstamas:
This page references some of the removed Ant files https://maven.apache.org/guides/getting-started/index.html

PR #459, opened 2023-10-03
fix style

imba-tjd:
I didn't learn apt syntax. I guessed it from the context.

I think this is not what you want

图片


slawekjaranowski:
Changed page: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html

slawekjaranowski:
@imba-tjd thanks

PR #464, opened 2023-11-02
MNGSITE-527 broken link to settings-1.3.0.xsd

froque:
JIRA issue: https://issues.apache.org/jira/browse/MNGSITE-527


gnodet (changes requested):
If we are to publish 4.0.0-alpha related XSDs, we'll also miss:
  * maven-4.1.0.xsd
  * core-extensions-1.2.0.xsd
  * plugin-1.1.0.xsd

froque:
Running `mvn verify` on latest maven master only creates a `settings-1.3.0.xsd`. How can I create those 3 files ?

ditbace:
Thanks

ditbace:
I don't know that is what I'm trying to do.

gnodet:
The two missing XSDs should now be available by building maven master branch.

gnodet:
The `plugin-[version].xsd` schemas are currently not published, so don't bother with those.

gnodet (approved):

gnodet:
I'm slightly worried about publishing unstable XSDs.  This may cause problem is the next 4.0.0 alpha/beta/milestone release does actually change it.  Users may have the schema cached by their IDE already.  A safe way would be to bump the version of the schemas if they changed in a future release, irrespective of its alpha status.  Another one would be to publish those with `-alpha-8` or maybe just `alpha` in their file name, so something like `maven-4.1.0-alpha-8.xsd`... ?

gnodet:
Note that adding `-alpha-8` in the file names won't exactly fix the existing broken links.
But at least, it provides an easy fix...

slawekjaranowski:
Should we have license headers in files? 
Does missing license will break site builds?

slawekjaranowski (changes requested):
```
mvn verify 

....

[WARNING] Files with unapproved licenses:
  content/resources/xsd/core-extensions-1.2.0-alpha-8.xsd
  content/resources/xsd/maven-4.1.0-alpha-8.xsd
  content/resources/xsd/assembly-2.2.0.xsd
  content/resources/xsd/settings-1.3.0-alpha-8.xsd
  content/resources/xsd/assembly-component-2.2.0.xsd
...

slawekjaranowski:
By the way ... maybe we should create html page at https://maven.apache.org/xsd/ with description for which component and version specific xsd should be used

another idea - add to xsd comments component and version

of course all for next PRs

gnodet:
> Should we have license headers in files? Does missing license will break site builds?

Those files are generated.  Not sure if the old ones have been manually edited to add the header before publication or if the generation somehow changed in modello...

> By the way ... maybe we should create html page at https://maven.apache.org/xsd/ with description for which component and version specific xsd should be used
> 
> another idea - add to xsd comments component and version
> 
> of course all for next PRs

I'd like to avoid maintaining the whole list of schemas manually.  I think using a `.htaccess` could be leveraged to provide a `readme.html` to display on top of the listing. See [apache mod_autoindex](https://httpd.apache.org/docs/2.0/mod/mod_autoindex.html#indexoptions)

michael-o (approved):

slawekjaranowski (approved):

jira-importer:
Resolve #1335

PR #482, opened 2023-12-31
Use puml to generate sources diagrams

slawekjaranowski:
We have diagram of Maven sources repositories: https://maven.apache.org/scm.html
https://maven.apache.org/maven-sources.png

I don't see a way to easy edit this png file .... so I try to use a puml for generate it.

The result can look like:

image

Links in puml also working 😄 

If you like it I will continue work.

### DIFF
--- a/src/plantuml/maven-sources/plugins.puml
+++ b/src/plantuml/maven-sources/plugins.puml
@@ -1,0 +1,116 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/
+@startuml
+!pragma layout smetana
+
+skinparam rectangle {
+  BackgroundColor LightBlue
+}
+
+
+package Plugins {
+
+  package Core {
+    rectangle clean
+    rectangle compiler
+    rectangle deploy
+    rectangle install
+    rectangle resources
+    rectangle site
+    rectangle surefire
+    rectangle verifier
+  }
+
+  package Packaging {
+    rectangle ear
+    rectangle ejb
+    rectangle jar
+    rectangle rar
+    rectangle war
+    rectangle acr
+    rectangle shade
+    rectangle source
+    rectangle jlink
+    rectangle jmod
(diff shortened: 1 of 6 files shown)


mthmulders:
Big +1 from my side for having a text source for these kind of diagrams, rather than a binary one.

src/plantuml/maven-sources/plugins.puml
```diff
@@ -0,0 +1,116 @@
...
+    rectangle invoker
+    rectangle jarsigner
+    rectangle jdeprscan
+    rectangle patch
+    rectangle pdf
+    rectangle plugin
+    rectangle release
+    rectangle "remote-resource"
+    rectangle scm
+    rectangle "scm-ppublish"
```
mthmulders, line 83: Typo, or is it really named "scm-ppublish" (double 'p')?

src/plantuml/maven-sources/plugins.puml
```diff
@@ -0,0 +1,116 @@
...
+    rectangle toolchains
+  }
+}
+
+'hidden dependencies to change layout
+Core      -[hidden]-> Packaging
+Packaging -[hidden]-> Reporting
+Reporting -[hidden]-> Tools
+
+clean     -[hidden]> compiler
```
mthmulders, line 95: My PlantUML knowledge is a bit rusty... Is there a difference between `-[hidden]->` (as above) and `-[hidden]>` (here)?

slawekjaranowski: I also learn by testing ...https://plantuml.com/component-diagram

> By default, links between classes have two dashes -- and are vertically oriented. It is possible to use horizontal link by putting a single dash (or dot) like this:

mthmulders: From the same page:

>  It is also possible to change arrow direction by adding `left`, `right`, `up` or `down` keywords inside the arrow: 

How would you feel about using that notation? I feel it might be friendlier to those who aren't so familiar with PlantUML and wish to make changes to this diagram.

hboutemy:
> I don't see a way to easy edit this png file

FTR, source OpenOffice document has been committed with output png in 9a8edf77fdaeba50615240670eae044b0a5be02b

but using PlantUML looks like a very good idea: trying on this schema as a first case is great. I'd love to be able to replace this OpenOffice approach to all the schemas I did, because it is a way to edit, but. not really easy

ideally also, generating an image map with links to target urls would be great

Bukama:
I talked to @slawekjaranowski and offered him to support him on the PlanUML issue (even having no experience in creating diagrams, but well sometimes is always the first time).

In my fork I started and I think I got the sources diagram okay and also the diagram for the multi subproject guide.

But I really struggle to create images like the Maven central picture, as plantUML doesn't offers (by design!) possibilities to "manually" draw lines :( So I'm not sure if it's worth trying to convert non-UML diagramms at all costs.

You can see my progress here: https://github.com/Bukama/MavenStuff/blob/main/Maven4/puml/pumlworking.md

What do you think @hboutemy @slawekjaranowski ?

slawekjaranowski:
@Bukama thanks for works we can finish this PR for source site ... another should be in separate PR

slawekjaranowski:
Suppressed by #615

PR #483, opened 2024-01-05
docck Maven Plugin was retired so remove reference from docs

slawekjaranowski:


gnodet:
It would be nice to indicate the reason why the plugin was retired.

gnodet (approved):

slawekjaranowski:
https://issues.apache.org/jira/browse/MDOCCK-39

PR #485, opened 2024-01-15
Clarify default values for repositories

kwin:
Clean up example snippet for repositories

### DIFF
--- a/content/apt/pom.apt.vm
+++ b/content/apt/pom.apt.vm
@@ -1710,12 +1710,10 @@ scm:cvs:pserver:127.0.0.1:/cvs/root:my-project
     <repository>
       <releases>
         <enabled>false</enabled>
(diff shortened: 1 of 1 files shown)


cstamas:
LGTM

Maybe, but just maybe add a sentence explaining this https://issues.apache.org/jira/browse/MRESOLVER-277 more specifically this comment https://issues.apache.org/jira/browse/MRESOLVER-277?focusedCommentId=17614487&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17614487

Here user had _empty_ repo and expected policy to kick in, that is applied only to cached content.

kwin:
It currently says:

> **never** (only downloads if not yet existing in the local repository)

For the this already covers the never != offline aspect.

PR #488, opened 2024-01-25
Update index.apt

pzrq:
Adds <<<>>> syntax around pom.xml, settings.xml and a few other instances which seem obvious to me.


hboutemy:
thank you

PR #489, opened 2024-01-25
Improve wording on artifact:compare introducing false positives

lucamolteni:


content/apt/guides/mini/guide-reproducible-builds.apt
```diff
@@ -78,7 +78,10 @@ mvn clean verify artifact:compare
 
  []
 
- Notice that this does NOT really prove that your build is yet reproducible, because your build may still suffer from environment leaks (username, current directory, ...). But it is easy to do, and prevents basic non-reproducible issues like timestamps.
+ Notice that this does NOT really prove that your build is yet reproducible, because your build may still suffer from environment leaks (username, current directory, ...). 
+ Also not that it might suffer from false positives. Libraries included in your build might not support reproducible builds themselves but at the same time produce reproducible artifacts.
```
yurloc, line 82: In my opinion, this is inaccurate and adds more confusion than it removes. The false positive check result we've seen in OptaPlanner didn't come from the `artifact:compare` check that this paragraph is about but from `artifact:check-buildplan`, which is mentioned in the next section.

@hboutemy I'd like to check with you first, before we try to polish this up and go into perhaps unnecessary details about false positives and the optionality of some checks... Do you think we could remove the jaxb2 plugin from the [issue list](https://github.com/apache/maven-artifact-plugin/blob/c0ab5135f9208b72e58a7b89bba48593a31674f1/src/site/apt/plugin-issues.apt#L73) since it turned out that using that plugin doesn't make the project build non-reproducible (only the plugin's build is not reproducible)?

yurloc: If that is true and we could remove it then the false-positive failure of `artifact:check-buildplan` would disappear and so would the confusion and this paragraph wouldn't need any update.

lucamolteni: In the previous section you mean. 

Besides you're right, I forgot which task triggered the problem and changed the wrong section but for `check-build plan` that still holds right?

slachiewicz:
Needs rebase

Bukama:
@lucamolteni Please update/rebase the PR, if you still want to contribute the changes. Thank you.

PR #494, opened 2024-02-25
Edit and update

elharo:
Focus the description on how this is used by the developer rather than by Maven.  Also, Java 5 is a little old, and sun is defunct. @hboutemy

### DIFF
--- a/content/apt/guides/mini/guide-using-toolchains.apt
+++ b/content/apt/guides/mini/guide-using-toolchains.apt
@@ -31,15 +31,13 @@
 
 Guide to Using Toolchains
 
-* What is Toolchains?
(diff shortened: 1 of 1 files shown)


slawekjaranowski (approved):

slawekjaranowski:
We should also update plugins version used in examples ... 
we can use properties from parent like in: https://github.com/apache/maven-site/commit/6d4a7bc25dcb491600ac3821481018d2be371003

slawekjaranowski:
Similar in plugin docs https://maven.apache.org/plugins/maven-toolchains-plugin/toolchains/jdk.html
Maybe we should drop a similar site from plugin documentations and have only one here ...

PR #504, opened 2024-03-18
use svg instead of png + imagemap

hboutemy:
makes updates much easier: hyperlink can be added to source .odf file
then export to svg (select all then limit export to selection) in root directory
then run `./prepare-svg.sh`

requires install of `svgo` to rework the svg exported by LibreOffice: https://svgo.dev/


slawekjaranowski:
What with: #482 - with puml looks to easier edit, no additional tools will be required
I simply must to find a time to finish a work ... or someone more can help

hboutemy:
@slawekjaranowski you're right, I forgot about it
for this simple sources list case, yes, puml can be a good solution
my objective is to rework https://maven.apache.org/ref/4.0.0-alpha-13/ : I don't think I'll be able to do it with puml
I started here because it is a simpler deployment setup, to test features and check it works as expected

hboutemy:
> I started here because it is a simpler deployment setup

Just after writing that sentence, I realize that testing on https://maven.apache.org/plugin-tools-archives/plugin-tools-LATEST/ is even easier to get old and new output at the same time

I'll do the test on plugin-tools, which better represent the type of full need

on this sources image and map, it's an opportunity to try to go ahead in parallel

.gitignore
```diff
@@ -8,3 +8,5 @@ test-output
 .idea
 .DS_Store
 .factorypath
+
+/*.svg
```
elharo, line 12: why are we ignoring svg? Isn't the point of this PR to include SVG files?

Bukama:
> my objective is to rework https://maven.apache.org/ref/4.0.0-alpha-13/ : I don't think I'll be able to do it with puml, unless you prove me wrong I started here because it is a simpler deployment setup, to test features and check it works as expected

I tried...well...looks..."wild" :(

image


P.S. relations exactly as on the current one, could be optimized a little bit, but well

elharo (changes requested):
The original is easier to read. Definitely make the font bigger and bolder. I'm trying to make up my mind about straight vs ones with right angles.

hboutemy:
replaced by #504

PR #509, opened 2024-04-11
Fix a typo.

algonell:
It seems that something is missing in the first sentence of this paragraph.

There are multiple options, but just adding "work" seems to be fine.

Thinking out loud, it could also be:

"Using archetypes provides a great way to enable developers create projects quickly in a way consistent with best practices employed by your project or organization."


hboutemy:
thank you

PR #518, opened 2024-05-01
Add hint about mvnrepository.com

tamaro-skaljic:
in the getting started with Maven in 30 minutes guide

### DIFF
--- a/content/apt/guides/getting-started/index.apt
+++ b/content/apt/guides/getting-started/index.apt
@@ -910,9 +910,9 @@ mvn process-resources "-Dcommand.line.prop=hello again"
  various plugins used to build the project).  By default, the remote repository Maven uses can be found (and browsed) at
  {{https://repo.maven.apache.org/maven2/}}.  You can also set up your own remote repository (maybe a central repository for your company) to
  use instead of or in addition to the default remote repository.  For more information on repositories you can refer to the
- {{{../introduction/introduction-to-repositories.html}Introduction to Repositories}}. The best way to find information about dependencies is
+ {{{../introduction/introduction-to-repositories.html}Introduction to Repositories}}. One aggregated way to find information about dependencies is
  {{{https://mvnrepository.com/}mvnrepository.com}}, where you can find not only all the groups and artifacts in Maven Central, but also those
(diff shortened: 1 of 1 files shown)


content/apt/guides/getting-started/index.apt
```diff
@@ -910,7 +910,9 @@ mvn process-resources "-Dcommand.line.prop=hello again"
  various plugins used to build the project).  By default, the remote repository Maven uses can be found (and browsed) at
  {{https://repo.maven.apache.org/maven2/}}.  You can also set up your own remote repository (maybe a central repository for your company) to
  use instead of or in addition to the default remote repository.  For more information on repositories you can refer to the
- {{{../introduction/introduction-to-repositories.html}Introduction to Repositories}}.
+ {{{../introduction/introduction-to-repositories.html}Introduction to Repositories}}. The best way to find information about dependencies is
```
rmannibucau, line 913: Maybe rephrase it to highlight it is not always in sync with central so maybe something like "one aggregated way to find dependencies.....but it can be outdated" rather than "the best way"?

tamaro-skaljic: Done.

elharo (commented):
I wouldn't include this. I find that site to be spammy and SEO clutter.

cstamas:
> I wouldn't include this. I find that site to be spammy and SEO clutter.

@elharo care to explain? What means "spammy and SEO clutter" in relation to https://mvnrepository.com/ ?

elharo:
It's never the site I want to see, but it's always there at the top of the search results, more often than not ahead of the project site for whatever I'm actually looking for. Also, it's been a few years so details are fuzzy, but at a previous company/team/project I recall I frequently had to explain to customers that the problems they were seeing were because they were on mvnrepository.com instead of where they should have been.

rmannibucau:
It is not the only one so guess it is either we enable any to be on our website or none.
Im for having them but mention their pitfalls (inaccurate results mainly) explicitly.
If it helps, all good for users, else just ignore IMHO.
Not worse than when the doc says to use a maven plugin, it stays a proposal but there are always better options for some people.

PR #569, opened 2024-10-23
Remove "Commit then Review"

elharo:
This is not what we're usually doing and not what we should be doing.

I avoided making more affirmat\ive statements about what we should be doing. I'm not sure this is the right page for that, but I'll look through the site and see if we say anything anywhere else.

### DIFF
--- a/content/markdown/project-roles.md
+++ b/content/markdown/project-roles.md
@@ -134,7 +134,7 @@ These are those people who have been given write access to the
 Apache Maven code repository and have a signed 
 [Contributor License Agreement (CLA)][4] on file with the ASF.
 
-The Apache Maven project uses a Commit then Review policy and has
+The Apache Maven project has
 [a number of conventions][5] which should be followed.
 
 Committers are responsible for ensuring that every file they 


michael-o (approved):
This does not imply the opposite, but simply that no process is formally set. I personally never likes CTR

ditbace:
ok

slawekjaranowski:
So what commit policy now we have? I think that should be discussed on ML and finish with formal vote.

To be clear I'm no a fun of CRT, I like reviews and see benefit of it, so on ...

michael-o:
> So what commit policy now we have? I think that should be discussed on ML and finish with formal vote.
> 
> To be clear I'm no a fun of CRT, I like reviews and see benefit of it, so on ...

Same here, but I don't see a need for a simply typo fix or a release. This can go straight to master.

olamy:
Agree this change should be reverted.
By the way there is not even approval or discussion here...

elharo:
I've seen too many questionable commits sneak in without review to be comfortable with it. It's not commit then review. It never has been. It's Commit Never Review. This is a massive security hole in the face of supply chain attacks like the one that almost compromised xz utils:

https://en.wikipedia.org/wiki/XZ_Utils_backdoor

I do not think it's necessary to be a big company with clear tasking to require advance code review. Any team project used as widely as Maven is must require code review.

hboutemy:
> this change should be reverted.

yes, ignoring what existed before GitHub Pull Requests and deciding unilaterally to change project practice is not an option (even with best intent of fixing a massive security hole, and even if this was done through an approved PR...)

but instead of just reverting, we should probably be smarter than the revert: state that no strict review is enforced, nor PR and approval, but best practices and judgement

Need to take time to write: help welcome

elharo:
Yes, we should make it explicit that review comes before commit and enforce that in the source repository. This hasn't happened yet, but we should discuss and do that. This won't be the first time Maven has had to change its practice to meet evolving security threats and sadly probably won't be the last.

olamy:
Personally I wouldn't make reviews required but at least PR required.
Some changes are trivial and doesn't require reviews and by the way people should have enough common sense to ask when they are in doubt.
I have seen so many fake reviews (how can people review code changes even very large in just a few minutes...)
What about having PR more enforced at least for traceability, release notes etc...

hboutemy:
before changing project policy, let's document our reality: just "CTR" is too crude as it was the pure reality when we were with svn = something from the past

let's write what we do with GitHub nowadays: promote PRs to help building release notes, and approval from peers when possible (and go direct when it makes sense)

on Git sub-projects that community feels are active enough, we can later to go stricter and activate main branch protection, like it seems this PR start with https://github.com/apache/maven-dependency-plugin/pull/444

step by step, not one size fits all change everywhere

PR #581, opened 2024-11-24
Fix link to Plugins section

arend-von-reinersdorff:
Link is currently broken, it links to #plugins but should be Plugins. See https://maven.apache.org/pom.html#:~:text=with%20the%20attributes%20combine.children%20or%20combine.self%20outlined%20in


slachiewicz (approved):

ditbace:
ok

PR #589, opened 2024-12-07
MNGSITE-546 Remove download cgi

kwin:


slawekjaranowski (approved):
Ok, let's try it

we still have redirects in: .htaccess

```
Redirect Permanent /download.html /download.cgi
RedirectMatch permanent ^/(.*)/download.html$ /$1/download.cgi
```
do we need it?

we have references in docs to `download.cgi`

kwin:
> we have references in docs to download.cgi

Not everywhere unfortunately, see for example https://github.com/apache/maven/blob/8d4f455ac9f29904bb2c86847d41c310782fbea6/src/site/site.xml#L46

jira-importer:
Resolve #1128

PR #595, opened 2024-12-16
#MNGSITE-539 Formatting the affected files

Bukama:
This PR formats the files that will later be updated with the actual content  for MNGSITE-539, so the formatting is split.


slawekjaranowski (approved):
I see only reformatings .... 😄

jira-importer:
Resolve #1125

PR #608, opened 2024-12-27
[MNGSITE-393] Remove Maven 2 Guide "Building for different environments with Maven 2"

Bukama:
@elharo asked me to create a PR to remove this Maven 2 guide, after I asked if this should also be removed.

image


content/apt/guides/mini/index.txt
```diff
@@ -39,8 +39,6 @@
 
  * {{{$prefix/guide-using-extensions.html}Using Extensions}}
```
elharo, line 41: Keep the link. Just remove the phrase "with Maven 2"

elharo (changes requested):

content/apt/guides/mini/index.txt
Bukama: Reverted and changed

elharo (approved):

PR #614, opened 2024-12-29
[MNGSITE-504] Fix rendering / markdown warnings

Bukama:
With this PR I updated some very old (, but still existing) pages to reduce the renderer warnings during the `site` goal (see MNGSITE-504).

This PR is **not* to update the outdated content of those pages!

I could fix all warnings in the existing files, except the following four:

```
[WARNING] Unrecognized xdoc tag <base> at [32:52]
[WARNING] Unrecognized xdoc tag <link> at [34:100]
[WARNING] Unrecognized xdoc tag <link> at [35:75]
[WARNING] Unrecognized xdoc tag <link> at [36:90]
```

These are from CSS definition in the `errors/404.xml.vm`. I couldn't find a documentation how to define them properly, esp. as we need dynamic links. This why I transformed the page to a markdown file.

edit: Reverted changes done to `404.xml.vm` - see discussion below.


content/markdown/errors/404.md
```diff
@@ -0,0 +1,29 @@
+<!--
```
elharo, line 1: I don't think markdown supports comments?

content/xdoc/errors/404.xml.vm
slawekjaranowski, line 1: We can not move it to `md`
We need additional html page headers.

This content can be displayed in any path .... and without those header content will be corrupted.

@kwin @michael-o can we add html headers in `md` format?

slawekjaranowski (changes requested):

content/xdoc/errors/404.xml.vm
Bukama: > We can not move it to `md` We need additional html page headers.
> 
> This content can be displayed in any path .... and without those header content will be corrupted.
> 
> @kwin @michael-o can we add html headers in `md` format?

To add as an information what I've tested locally:
I checked the `.htaccess` which redirects to this in case of 404 and I checked if `/errors/404.html` is displayed correctly - which it is.
But I'm honest: I'm not very experienced with `.htaccess` file (as you probably already noticed).

content/markdown/errors/404.md
Bukama: > I don't think markdown supports comments?

Works fine:
image

elharo: I think that's a function of some particular implementation that extends the spec. I'm not sure we can/should rely on that.

Bukama: > I think that's a function of some particular implementation that extends the spec. I'm not sure we can/should rely on that.

Then we need to change **all** other `.md` files too and remove every comments and licences there...

content/xdoc/errors/404.xml.vm
slawekjaranowski: The problem will be on page like: https://maven.apache.org/plugins/maven-abc-plugin/

content/markdown/errors/404.md
michael-o: No, that is a bug @kwin had ready fixed.on master.

elharo: My bad. I forgot about inline HTML. This is OK synatctically

content/xdoc/errors/404.xml.vm
Bukama: Okay, will wait for an answer of @kwin or @michael-o about your question and then change or revert the 404 file

kwin: In general every valid HTML is also valid in MD (https://daringfireball.net/projects/markdown/syntax#html). This is more a question on how the resulting HTML is being partitioned by the doxia-sitetools (to inject things like navigation).

Bukama: > In general every valid HTML is also valid in MD (https://daringfireball.net/projects/markdown/syntax#html). This is more a question on how the resulting HTML is being partitioned by the doxia-sitetools (to inject things like navigation).

Thank you for your response. For the moment I reverted the changes in the 404 (meaning the `vm` is back there and the `md` is gone), until we are sure. To change the 404 to MD is done in few mins anyway.
But this means that the rendering warnings, caused by this page (double title and the HTTP `Unrecognized xdoc tag` are back in place)

Bukama: Side note: Did some tests about using test and filed an issue [DOXIASITETOOLS-358](https://issues.apache.org/jira/browse/DOXIASITETOOLS-358)

Bukama:
Rebased after several other commits were merged into master

jira-importer:
Resolve #1110

PR #616, opened 2024-12-29
Convert FML files into Markdown

Bukama:
Similar to converting the APT files to Markdown (MNGSITE-507 / #573), this PR converts the existing FML files.

Note: It's not intended to update content with this PR, but only format, so content updates are clearly visible in diffs later.


slawekjaranowski:
By the way `about.html` is not referenced anywhere ...

slawekjaranowski (approved):

PR #621, opened 2025-01-02
Convert xdoc (without parsing) to Markdown

Bukama:
After by tests of HTTP meta properties (see results [DOXIASITETOOLS-358](https://issues.apache.org/jira/browse/DOXIASITETOOLS-358)) I converted all of those xdoc files to Markdown where no parsing/calculation happens. We can do this when we have a parser for this.

**Please note**

General
* HTTP standard about meta tag for author (`<meta name="author" content="name of author">`) does not define an attribute for the email of the author. The content of those (like in `mojo-api-specification.xml` is not converted regardless of xdoc or markdown).
 

In the `articles.md`:

I did the following changes
* updated or removed links to buy the books,
* updated the link to the video
* changed the divs to a table for a more similar layout and because inside of divs not markdown is possible and I didn't want to copy all HTML.

I did **not**
* updated the articles list, even that there a tons of not-exists-anymore or unsafe (http instead of https) links, 
* added the text and video of Karl Heinz (see #324)

Simply because I realized that it's far more than transformation then. I would like to update the list and add Karl Heinz's video and other articles and videos to the list in a separate PR (pr an updated 324).

In the `mojo-api-specification.md`

* I moved the sentence about Maven 4 to the top, similar to the "multi module guide" and added Maven 3 to the tile
* I did not converted the commented out code, which is not rendered by the xdoc converter at the moment too, so no content loss here

In the `repository.index.md`
I did copy the image map, because I'm not happy yet with my plantUML image. Therefore I also moved the `.odg`files.

In the `scm.md`
I added table headers and created own tables for each section (on live there are partly filled table rows used as `in between headers`).


Bukama:
Converted to draft, because I remembered that I wanted to add the JIRA links for the plugins (instead)

Bukama:
Added issue links and rebased after master's changes. Ready for review.

content/markdown/scm.md
```diff
@@ -0,0 +1,226 @@
...
+   repo init -u https://gitbox.apache.org/repos/asf/maven-sources.git`
+   repo sync
+   repo start master --all
+   ```
+3. In your IDE, import the projects you're interested in from the repo workspace.
+   Or directly build with command line the component you want.
+
+## Maven Sources Overview
+
+![Maven Sources: Site](maven-sources/site.svg)
```
slawekjaranowski, line 58: Next issue for markdown ... in version generated from xdoc we have `<object ...>` in output html.
With `img` tag links in svg are not working ...

@kwin @michael-o can you look?

slawekjaranowski: ok, I see we can add a html code to markdown ... so should be 
`<object ..../>` as in old document version

Bukama: Updated:
Note Manual <p> tags needed to align the images below each other. Without them they were all next to each other

slachiewicz (approved):
Good to go, we can improve later

PR #626, opened 2025-01-04
Require Java 17

slachiewicz:


slawekjaranowski (approved):

pom.xml
```diff
@@ -144,54 +144,6 @@
...
-          <execution>
-            <id>enforce-java8</id>
-            <goals>
-              <goal>enforce</goal>
-            </goals>
-            <phase>site</phase>
-            <configuration>
-              <rules>
-                <requireJavaVersion>
-                  <version>1.8</version>
```
Bukama, line 168: Why not keep the enforcer and also update it to require at least 17?

Bukama (approved):
+1 (nb)

pom.xml
slachiewicz: Enforcer rule is already defined in parent

PR #627, opened 2025-01-04
Apply Spotless to Markdown files

slachiewicz:


Bukama:
Yes please :D

Only comment so far, because builds fails due spotless errors :D But I guess you have already seen that an already working on it :)

slachiewicz:
sometimes, spotless breaks content - the solution is to add a new line before paragraph. 

>  -Apache·Maven·2.x·has·reached·its·end·of·life·and·is·no·longer·supported.
        +Apache·Maven·2.x·has·reached·its·end·of·life·and·is·no·longer·support-->

slawekjaranowski (approved):
lets go 😄

PR #659, opened 2025-01-31
Add example of profile activation with value true

arend-von-reinersdorff:
I was looking for an example like this recently. And I was surprised that there's an example for profile activation with property value false but not example of profile activation with property value true.

### DIFF
--- a/content/apt/guides/introduction/introduction-to-profiles.apt
+++ b/content/apt/guides/introduction/introduction-to-profiles.apt
@@ -260,6 +260,30 @@ mvn groupId:artifactId:goal -P profile-1,profile-2,?profile-3
 </profiles>
 +---+
   
+  The following profile will be activated when the system property "debug" is defined with no value, 
+  or is defined with the value "true".
+
++---+
+<profiles>
+  <profile>
+    <activation>
+      <property>
+        <name>debug</name>
+        <value>true</value>
+      </property>
+    </activation>
+    ...
+  </profile>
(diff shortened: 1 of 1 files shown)


content/apt/guides/introduction/introduction-to-profiles.apt
```diff
@@ -260,6 +260,30 @@ mvn groupId:artifactId:goal -P profile-1,profile-2,?profile-3
...
+
++---+
+<profiles>
+  <profile>
+    <activation>
+      <property>
+        <name>debug</name>
+        <value>true</value>
+      </property>
+    </activation>
```
slawekjaranowski, line 274: We have already example like:
```
    <activation>
      <property>
        <name>environment</name>
        <value>test</value>
      </property>
    </activation>
```

is it not the same?

arend-von-reinersdorff: value=true is a special case which allows to set the property on the command line without a value like this:
mvn groupId:artifactId:goal -Ddebug

Other values require a value to be specified on the command line like this:
mvn groupId:artifactId:goal -Ddebug=test

slawekjaranowski: Ok, but it is not a feature o profile but feature how user properties are initialized by CLI

By the way we need check if it documented somewhere

slawekjaranowski (approved):

PR #662, opened 2025-02-05
Add dependencies page

elharo:
This is a more careful and accurate definition of dependencies than we have in the POM reference or on the artifacts page. In particular, it is careful to distinguish between dependencies and artifacts. That the POM reference often conflates these two different concepts has confused me more than a little.

### DIFF
--- a/content/markdown/repositories/dependencies.md
+++ b/content/markdown/repositories/dependencies.md
@@ -20,7 +20,7 @@ under the License.
 -->
 
 A dependency is a connection between a Maven project and an artifact.
-Maven will add the artifact to one or more of the project's classpaths 
+Maven will add the artifact to one or more of the project's classpaths
 or otherwise use it when building the project.
 A dependency includes the artifact coordinates (group ID, artifact ID, and version)
 and metadata about how the artifact is used by the project. This metadata includes


elharo:
Flaky CI:

ERROR: Error cloning remote repo 'origin'
hudson.plugins.git.GitException: Could not init /home/jenkins/jenkins-home/workspace/aven_maven-box_maven-site_PR-662@libs/

elharo (commented):
Ping. I'd like to move this one along. Thanks.

slawekjaranowski (commented):
Is it not a duplicate somehow with https://maven.apache.org/repositories/artifacts.html
Should we also add in menu?

elharo:
No, it is absolutely not a duplicate. Artifacts and dependencies are not the same thing, and straightening that out is why I wrote this (as well as rewriting the page you reference).

elharo:
I added it to content/markdown/repositories/index.md

is there another index page you'd also like this added to?

gnodet:
> I added it to content/markdown/repositories/index.md
> 
> is there another index page you'd also like this added to?

We should maybe somehow align with https://maven.apache.org/ref/4.0.0-rc-2/apidocs/org/apache/maven/api/package-summary.html

elharo:
Thanks for pointing me at that. I think I need to take a serious editing pass across that. I also see some mismatches in the API, but I need to think a little more deeply about that. I don't think this PR needs to wait for any of that to happen though.

content/markdown/repositories/dependencies.md
```diff
@@ -0,0 +1,232 @@
...
+
+<!-- As I write this, I'm realizing that Maven overloads dependency scope
+for multiple different purposes. In particular we've confused the scope in the
+classpath with where the artifact lives and how it's found.
+Scope should be one of compile, runtime, test, or all.
+system should be replaced by systempath. provided and import should be separate elements.
+Or perhaps we should have a separate source element that has values
+repository, system (or url),  provided, and import.
+Not that we're going to fix any of this now, but it is worth
+understanding why this is confusing.
```
gnodet, line 140: It may be worth looking at the Maven 4 API which has tried to make things more obvious.
There's a `DependencyScope` enum which actually has a `compile-only` value.
And when resolving class/module paths, we end up with `Map<PathType, List<Path>>` and not a simple list of paths anymore.

content/markdown/repositories/dependencies.md
```diff
@@ -0,0 +1,232 @@
...
+| pom          | `pom`     |              |
+| maven-plugin | `jar`     |              |
+| ear          | `ear`     |              |
+| ejb          | `jar`     |              |
+| ejb-client   | `jar`     | `ejb-client` |
+| javadoc      | `jar`     | `javadoc`    |
+| java-source  | `jar`     | `sources`    |
+| rar          | `rar`     |              |
+| test-jar     | `jar`     | `tests`      |
+| war          | `war`     |              |
```
gnodet, line 184: We have a bunch of additional new types, see https://maven.apache.org/ref/4.0.0-rc-2/apidocs/org/apache/maven/api/Type.html

content/markdown/repositories/dependencies.md
elharo: Yes, I'll take a look. That said, this page is written for Maven users who primarily interact with all this through pom.xml files and never touch the API. This needs to be compatible with the API terms, but shouldn't expect knowledge of the API from the reader.

content/markdown/repositories/dependencies.md
elharo: Ok. We'll need to update this table when 4.0 is released, or maybe mark them 4.0 only here

gnodet:
> Thanks for pointing me at that. I think I need to take a serious editing pass across that. I also see some mismatches in the API, but I need to think a little more deeply about that. I don't think this PR needs to wait for any of that to happen though.

Sure.  Be aware that the dependency in the page I pointed at is the _resolver concept_, not the _POM concept_.  The _dependency_ described here is more linked to the _dependency coordinates_.  So don't think the terminology is wrong, it just cannot be mapped directly to the resolver API which is more complex.

gnodet (approved):
LGTM.
We need to keep in mind that this will slightly be out of date with Maven 4.

gnodet:
> > Thanks for pointing me at that. I think I need to take a serious editing pass across that. I also see some mismatches in the API, but I need to think a little more deeply about that. I don't think this PR needs to wait for any of that to happen though.
> 
> Sure. Be aware that the dependency in the page I pointed at is the _resolver concept_, not the _POM concept_. The _dependency_ described here is more linked to the _dependency coordinates_. So don't think the terminology is wrong, it just cannot be mapped directly to the resolver API which is more complex.

Fwiw, I think this page will have to be adjusted to a more fine grained terminology.  People do refer to "my project dependencies", or "this artifact dependencies" and actually refer to the transitive artifacts which are dependencies of a given artifact / project.  In the POM, you have (potentially incomplete) pointers to dependencies, and those are _dependency coordinates_, which will be turned into actual _dependencies_, once the resolver has _collected_ them (_resolution_ is the process of actually _downloading_ the artifacts).

PR #667, opened 2025-02-05
[MNGSITE-393] Repository layout hasn't changed since Maven 2

elharo:


hboutemy:
i don't get what we win by hiding history of Maven 1 and 2

elharo:
Consider the audience. Anyone coming to this page wants to know the current info, not what happened 10+ years ago that is no longer in production anywhere.

content/markdown/repository/layout.md
```diff
@@ -19,7 +19,9 @@ specific language governing permissions and limitations
 under the License.
 -->
 
-Maven2 repository layout format is the default layout used since Maven 2 (see [Repository Layout- Final](https://cwiki.apache.org/confluence/display/MAVENOLD/Repository+Layout+-+Final) in [Maven 2.0 Design Documents](https://cwiki.apache.org/confluence/display/MAVENOLD/Maven+2.0+Design+Documents)), superceding old Maven 1 layout:
```
hboutemy, line 22: perhaps even adding a link to Maven Central for Maven 1 http://repo1.maven.org/maven/
vs Maven Central for Maven 2 https://repo1.maven.org/maven2/
could help make the 2 very concrete

elharo: Do you know anyone still using Maven 1? (I find it ironic that in one PR I'm arguing against requiring Java 17 instead of Java 8, and in another I'm arguing to drop Maven 1.)

hboutemy: ok, I see: not listening

I won't discuss, seems to be lost time: just clarifying that I veto this PR

Bukama: To move forward with this PR, I suggest the following: We add a "Former/Historical Maven 1 layout" at the end of the page and put a one liner and the link from above down there. So the article focus on current layout, but also provides the historical information, but not that present as it is now.

slachiewicz (approved):

PR #672, opened 2025-02-08
Enforce Java17 or above for building

kwin:
This requirement has been introduced with the update of plantuml-maven-plugin to 0.3


kwin:
see also https://github.com/mthmulders/puml-maven-plugin?tab=readme-ov-file#requirements

slawekjaranowski (approved):

PR #675, opened 2025-02-08
Add support for 4.x suffix in doc helper

slawekjaranowski:


Bukama (approved):
It does work.

However, when I tested it locally I was confused that 
* clicking the v4 box does not executions `instructions()` (Loaded the page, then was selecting and deselecting the checkbox but the version suffix was and stayed gone from the instructions)
* switching between categories that are possible v4 ready the checkbox was reset to "false" (so you have to tick it and press the publish button again).

As I have never made a release myself yet, I'm not aware if these two things are important at all.

Aside from that I guess this page will be updated somewhat later, when the majority is v4 ready to v4 box is checked by default. To we want to create an issue for this as a reminder?

slawekjaranowski:
> It does work.
> 
> However, when I tested it locally I was confused that
> 
> * clicking the v4 box does not executions `instructions()` (Loaded the page, then was selecting and deselecting the checkbox but the version suffix was and stayed gone from the instructions)
> * switching between categories that are possible v4 ready the checkbox was reset to "false" (so you have to tick it and press the publish button again).
> 
> As I have never made a release myself yet, I'm not aware if these two things are important at all.
> 
> Aside from that I guess this page will be updated somewhat later, when the majority is v4 ready to v4 box is checked by default. To we want to create an issue for this as a reminder?

- next issue as reminder is not needed
- this is a tool which help to generate command which must be done manually by release manager when publishing new version of documentation

PR #680, opened 2025-02-09
Add hint about encryption of master password

kwin:


content/apt/guides/mini/guide-encryption.apt
```diff
@@ -83,6 +83,9 @@ mvn --encrypt-master-password <password>
 </settingsSecurity>
 +------------------------------------+
 
+ <The encrypted version of the master password is encrypted with a hardcoded key, so please treat it as if the password is stored in the file in plain text
```
elharo, line 86: Where is this hardcoded key stored? Is it per installation? per user? All Maven projects everywhere?

Period after plain text, and remove the parentheses.

kwin: The key is hardcoded in code: https://github.com/apache/maven/blob/fe25a2627c1dafeb44188dad9f45dfd5fe965a98/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java#L856

content/apt/guides/mini/guide-encryption.apt
```diff
@@ -83,6 +83,8 @@ mvn --encrypt-master-password <password>
 </settingsSecurity>
 +------------------------------------+
 
+ <The encrypted version of the master password is encrypted with a {{{https://github.com/apache/maven/blob/fe25a2627c1dafeb44188dad9f45dfd5fe965a98/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java#L856}hardcoded key}}, so please treat it as if the password is stored in the file in plain text.>
```
elharo, line 86: Is the ky in DefaultPlexusCipher? I don't see it?

elharo, line 86: strange, I don't even see a two-argument version of encryptAndDecorate in DefaultPlexusCipher. Are we using an older version of plexus?

kwin: First is the value to be encrypted, second the password to use for encryption

kwin: I don't think this PR is the proper place to discuss impl details, but you are looking at the wrong version of it. Maven 3.x uses 2.0 (https://github.com/apache/maven/blob/fe25a2627c1dafeb44188dad9f45dfd5fe965a98/pom.xml#L144): https://github.com/codehaus-plexus/plexus-cipher/blob/plexus-cipher-2.0/src/main/java/org/sonatype/plexus/components/cipher/PlexusCipher.java

kwin: Maven 4 got rid of it: https://maven.apache.org/guides/mini/guide-encryption-4.html.

elharo (approved):

PR #701, opened 2025-02-22
Add info about EOL to release history.

Bukama:
Based on the slack discussion, I added a small information what does EOL means for Maven and that there is no difference between EOL and EOS for us.

### DIFF
--- a/content/markdown/docs/history.md.vm
+++ b/content/markdown/docs/history.md.vm
@@ -131,7 +131,7 @@ $esc.h$esc.h Maven 3.8.x+
 
 $esc.h$esc.h Maven 3.6.x and before
 
-<p>Maven before 3.6.3 has now reached its end of life and is no longer supported in any way.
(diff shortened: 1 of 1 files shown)


content/markdown/docs/history.md.vm
```diff
@@ -126,7 +131,7 @@ $esc.h$esc.h Maven 3.8.x+
 
 $esc.h$esc.h Maven 3.6.x and before
 
-<p>Maven before 3.6.3 has now reached its end of life.
+<p>Maven before 3.6.3 has now reached its end of life and is no longer supported in any way.
```
slawekjaranowski, line 134: I tink that  3.6.3 also should be EOL ... so maybe 

> Maven 3.6.3 and before ...

as in line 132

Bukama: Good spot! Changed

slawekjaranowski (approved):

PR #702, opened 2025-02-22
[MNGSITE-441] More carefully define version string and token ordering

elharo:
No longer uses the ambiguous term "lexicographic". More clearly distinguish the comparison of version strings and the individual tokens parsed from the string. 

This is not intended to change anything, simply to more carefully document the existing behavior.


Bukama (approved):
Thanks for this additional update.
When merging, please change the title to `MNGSITE-441` according to the issue.

jira-importer:
Resolve #1004

PR #721, opened 2025-03-15
Use jenkins workspace as site cache

slawekjaranowski:
To avoid use old site cache from previous builds


hboutemy:
if  `.repository`  is cached, it may become corrupted as we have with the directory in user home: do you have an easy way to clean it?

just putting in `target/maven-site` with cost svn checkout on each build, but this will avoid corruption as it is never kept

slawekjaranowski:
> if `.repository` is cached, it may become corrupted as we have with the directory in user home: do you have an easy way to clean it?
> 
> just putting in `target/maven-site` with cost svn checkout on each build, but this will avoid corruption as it is never kept

ok, now I use target directory

hboutemy (approved):

PR #728, opened 2025-03-16
Convert apt to markdown

slachiewicz:


Bukama:
I like getting progress here as I still wait for Konrads announced updates in his PR (didn't want to just take over, after he announcced he will update). However I'm a bit curious about yours, based on your comment

> Let's open a few smaller, less problematic PR with some parts that could be easily reviewed and quickly merged. One big may be outdated in a few days

What did change your mind?

kwin (changes requested):
Please use the feature introduced with https://issues.apache.org/jira/browse/DOXIATOOLS-90 to use 2 commits and therefore the full history.

slachiewicz:
I've used instructions from https://github.com/apache/maven-site/pull/573#issue-2621918965 but forgot about this switch.

You are right - we should keep git history.

Issue #726, opened 2025-03-16
Some minor suggestions for improved navigation


raupachz:
### Improvement proposal

I'd like to suggest some minor refinements for a cleaner, more intuitive user experience. Opinionated, yes, I am fully aware.

**Header**

* Remove the "Apache Maven Project" image from the top left
* Move the "Maven" logo from the right to the left
* Drop the Breadcrumps, the two links "Download" and "Get Sources" and move last published to the footer line.

I would even go for the Maven Owl - it's cute - not sure if that is the correct name of the Maven mascot, or if it is even the Maven mascot, but I think the Maven logo is more well known. Anyways, these changes would significantly reduce visual clutter. I don't think the removed stuff really adds anything to the site. I don't want to be rude, offend anyone, just giving a personal opinion.

**About Maven - left site navigation**

* Rename "What is Maven?" to "Introduction". This matches the heading of the linked site.
* Move "What's new in Maven 4?" right below and rename it to "Journey to Maven 4"
* Remove "Maven Daemon" from this section.

I am not a fan of questions in site navigation. Also "Journey to Maven 4" makes it clear, that Maven 4 is not generally available yet - and it takes much time and effort to get to it. Maven Daemon is a Maven project and fits better in the other section.

**Maven Projects - left site navigation**

* Rename "Maven" to "Maven 3"
* Add a new item  "Maven 4" below and link to the to the Maven 4 project site
* Move the removed "Maven Daemon" here as "Daemon".

This makes it easier to understand that there is Maven 3 and Maven 4. One is the current one and the other one is the road ahead. Maven Daemon is moved where it belongs. I think calling it "Daemon" here is good enough. Everything is Maven in the end: plugins, archetypes...

I considered submitting these changes as a PR,  but wanted to get your opinions first. I understand these changes are tricky to implement since many Maven project sites share the same template and I do know know the ins and outs of site generation.

Feel free to give feedback or close this issue if you disagree. Just trying to help a bit.

Bukama:
Hi @raupachz ,
thank you for the issue!

Since (about three months now) I'm updating more and more outdated things on the site, I get lost on the page even more (as I look at more pages than as a regular user) and I don't think only renaming some entries will help users to find what they are looking for. Aside my current tasks, I was already thinking about a more deep improvement, but I'm have not gathered all thoughts (and their pros/cons) nor written them down (and will not before April/May due to lot of regular work and upcoming conferences). I will mention you, when I post them somewhere.

[...]

PR #737, opened 2025-03-19
Maven concepts

raupach-e2n:
Added a page Apache Maven Concepts to the top left navigation. Starting point for terms and concepts that are needed to understand how Apache Maven works. 

Nothing here is new. It is collection from other pages, that are hidden several pages below Documentation, User Guide and so on. 

This is a draft. 

Please let me know if this should be discussed on the mailing list.

### DIFF
--- a/content/markdown/concepts.md
+++ b/content/markdown/concepts.md
@@ -1,0 +1,34 @@
+# Apache Maven concepts
+
+The following terms and concepts are central to understanding and use of Apache Maven:
+
+* POM
+* Lifecycle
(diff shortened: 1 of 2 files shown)


Bukama (commented):
My thoughts:

1) These are not concepts, but some kind of glossary entries - we have a glossary page
2) We already have these information, but with more details in each of their articles. Maybe they are not that visible (A thing I also want to change, see my comment in #726)  

Therefore, I personally don't want to accept this PR.

PR #741, opened 2025-03-23
Moved Run and Configure to top navigation

raupachz:
Visitors to Apache Maven website are looking for information to install, run and configure maven.

Run and Configure should be visible and placed in top left site navigation. Currently it is hidden in the Use dropdown menu. I don't think this dropdown adds much value instead it hides important information.

The other two pages IDE Integration and Support and Training can be removed as well.

IDE Integration could be put under Installation, can be discussed, but I do not think it is worth mentioning anymore that IDEs support Apache Maven. Visual Studio Code among others isn't even mentioned, but does support Apache Maven. Apache Maven integration isn't such a novelty anymore that  is worth being mentioned.

Training and Support can be removed as well. Not saying that the mentioned companies are not worth your money, just that there is already so much coaching, training, YouTube videos and whatnot around that there is no need mentioning three companies.


slawekjaranowski:
@raupachz thanks for proposition ... can you split your PR to 3 separate ones - menu, ide, support

Bukama:
@raupachz Hey Björn, would you do the split please as requested by Slawek? Thank you :)

Bukama:
Split the PR into three single ones as requested by Slawek.
Therefore, closing this PR.

Thank you for your suggestion @raupachz !

PR #742, opened 2025-03-23
Downloads instead of Download

raupachz:
Minor change: It should say Downloads instead of Download in site navigation. 

There is more than one thing to download: different files for the latest Apache Maven version, Apache Maven Daemon (mvnd) and other previous stable releases.


Bukama (approved):

Bukama:
Thank you!

PR #752, opened 2025-04-05
Add all and each phase to Maven 4 article

Bukama:
Also small rephrasing to the recommendation to not use clean install for regular builds.


gnodet (commented):
Proposed  improvements, especially for `each` / `all`.

gnodet (approved):

PR #806, opened 2025-05-31
Maven 3.9.10 - release notes

slawekjaranowski:


content/markdown/docs/3.9.10/release-notes.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Overview About the Changes
+
+Regression fixes and other improvements from Maven 3.9.9. All users already on Maven 3.9.x are advised to upgrade.
+
+The majority of fixed issues are bug and regression fixes for user reported problems.
+
+This release updates Resolver to version [1.9.23](https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12320628&version=12354989).
+
+Maven itself received notable improvements in area of support for `JDK 24` as runtime.
```
Bukama: I'm afraid if users get this wrong and think Maven 3.9.10 needs Java 24 now (which makes no sense, but well)

content/markdown/docs/history.md.vm
```diff
@@ -104,7 +104,8 @@ $esc.h$esc.h Maven 3.8.x+
 <!--  release( date version announce latest-minor jdk rowspan ) -->
 <!-- The following two lines needs to be fixed in x days if the archive picks up the announcement mail. -->
 
-#release( "2024-08-17" "3.9.9" "https://lists.apache.org/thread/ov40dkfd1snyz2pdltsr5rzhdh2xd9yh" "true" "Java 8" "10" )
+#release( "2025-06-01" "3.9.10" "https://lists.apache.org/thread/to-fix" "true" "Java 8" "11" )
```
Bukama, line 107: Selfmemo: Placeholder for final URL

content/markdown/docs/3.9.10/release-notes.md
olamy: I agree this needs more details. I'm using jdk 24 with 3.9.9 to build Jetty 12.1.x for long time and there is no issue.  So reading this I would be surprised :)

slawekjaranowski: Ok, there was a many dependencies updates to be more compatible with JDK 24, like sisu and asm, 
some of warrnings on JDK 24 also was mutted or removed by updates - like for guava
so I think it is importants ... 
any suggestion how to mention about its in release notes.

Bukama: > Ok, there was a many dependencies updates to be more compatible with JDK 24, like sisu and asm, some of warrnings on JDK 24 also was mutted or removed by updates - like for guava so I think it is importants ... any suggestion how to mention about its in release notes.

Why note just tell exactly this in the release notes?

Maybe like this (including the disclaimer that only Java 8 is needed for those, who don't check history page"):

> When running a build using Java 24 with an earlier version of Maven there were a lot of errors and warnings, because many dependency were not compatible with a Java 24 runtime. In Maven 3.9.10 those got updated, resulting in Maven 3.9.10 now has a far better support if you want to run your builds on Java 24.
> 
> Note: Maven 3.9.10 still only requires Java 8 to execute.

slawekjaranowski: updated

Bukama (approved):

olamy (approved):

PR #824, opened 2025-06-18
Revert "Use next parent snapshot version"

elharo:


elharo:
Because 46-SNAPSHOT failed the build

PR #829, opened 2025-06-25
Edit "What is Maven?" for clarity

elharo:
This tries to eliminate some more awkward and jargony language. I also removed one phrase that I think is not true. That is, Maven does not generate code coverage reports out of the box.

### DIFF
--- a/content/markdown/what-is-maven.md
+++ b/content/markdown/what-is-maven.md
@@ -31,13 +31,12 @@ share JARs across several projects.
 The result is a tool that can now be used for building and managing any
 Java-based project. We hope that we have created something that will
 make the day-to-day work of Java developers easier and generally help
-with the comprehension of any Java-based project.
+them understand any Java-based project.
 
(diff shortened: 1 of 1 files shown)


content/markdown/what-is-maven.md
```diff
@@ -66,11 +65,10 @@ For example, Maven can provide:
 - Cross referenced sources
 - Mailing lists managed by the project
 - Dependencies used by the project
-- Unit test reports including coverage
```
olamy, line 69: Jacoco plugin does so why remove that?

elharo: Yes, and there are others but none are part of the core Maven project.

olamy: This is related to what Maven is doing in general and how it is extensible not only to what is maintained here.
We must talk about the vast community around Maven and everything you can do with it.
-1 for this change please revert this is a veto

Issue #811, opened 2025-06-07
Move all tools to "tools" (and add dedicated page below them)


Bukama:
### Improvement proposal

(From a discussion with @hboutemy)

Currently we have a dedicated ["tools" page](https://maven.apache.org/tools/index.html) with the goal to list all Maven tools there. However for [Daemon a separate page](https://maven.apache.org/mvnd.html) with menu entry was added.

This issue aims to 

* List "all" tools on the tools page
* Add a dedicated page for each tool "beneath" the tools one, where further information are provided (like the one for Daemon)

Like this:
```
- Maven Plugins
- Maven Extensions
- Maven Tools
  |- Daemon
  |- Wrapper
  |- ...
```

Tools to cover:
* Daemon
* Wrapper
* MvnSh(ell)
* Maven 4 Migration tool (#810)
* ???

PR #832, opened 2025-06-29
Add page about Maven Wrapper to tools section

Bukama:
Part of #811

### DIFF
--- a/content/markdown/tools/index.md
+++ b/content/markdown/tools/index.md
@@ -27,9 +27,11 @@ The pages in this section will give information about additional Maven tools, wh
 
 ## Supported By The Maven Project
 
-| Name                        | Description                                                                                      | Link        |
-|:----------------------------|:-------------------------------------------------------------------------------------------------|:------------|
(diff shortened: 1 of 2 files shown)


sparsick (approved):

hboutemy:
why creating a new page https://maven.apache.org/tools/mavenwrapper.html
when wrapper has a main page https://maven.apache.org/wrapper/

is there a strategy?
or just trying to be consistent with mvnd https://maven.apache.org/tools/mvnd.html?
hint: mvnd does not have a website associated to it, it's the only exception

Bukama:
> (joining late to the discussion, but the discussion is IMHO useful)
> 
> why creating a new page https://maven.apache.org/tools/mavenwrapper.html when wrapper has a main page https://maven.apache.org/wrapper/
> 
> is there a strategy? or just trying to be consistent with mvnd https://maven.apache.org/tools/mvnd.html? hint: mvnd does not have a website associated to it, it's the only exception

To be consistent with daemon, and the others I will write (and because I didn't know wrapper has documentation at at all)

PR #835, opened 2025-06-29
Move run and configure to top level

Bukama:


Bukama:
Moved into a own PR as requested in #741.

PR #1350, opened 2025-07-24
add namespace to examples

elharo:


elharo:
That's a separate issue. I'm personally not sure whether that's a good idea here or not. Might be, but if it is that's another PR.

elharo:
Also in many cases here the modelVersion is inferred in the ... that follows the project element

PR #1359, opened 2025-08-04
Convert gettingStarted/introduction from APT to Markdown

Bukama:


sparsick (approved):
I checked, if the converted pages are generated correctly

PR #1361, opened 2025-08-04
Convert remaining pages from APT to Markdown

Bukama:
Also fixing some formatting things in other migrated pages


sparsick (approved):
I check if the converted pages are generated correctly

Issue #1357, opened 2025-08-04
Clarify Dependency Scope table


brkeh:
### Improvement proposal

Dependency Scope table on page 

> Introduction to the Dependency Mechanism

Can we label the column and row headers? It took me a while to understand the table from the sentences above and I think header labels would it make clearer and easier to understand. My understanding is each row header in left column is a Main project dependency and each column header in first row is a Dependency project dependency.

I tried various formats in apt but HTML never rendered very well for me, too much space in some columns. Maybe something like this?

`*------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+`
`|                              | Main project dependency      |                              |                              |                              |
*------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+
|Dependency project dependency |                              | compile                      | provided                     | runtime                      | test
*------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+
|                              | compile                      | compile(*)                   |    -                         | runtime                      |  -
*------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+
|                              | provided                     | provided                     |    -                         | provided                     |  -
*------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+
|                              | runtime                      | runtime                      |    -                         | runtime                      |  -
*------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+
|                              | test                         | test                         |    -                         | test                         |  -
*------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+------------------------------+`

Bukama:
This is the table mentioned: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope

<img width="1600" height="277" alt="Image" src="https://github.com/user-attachments/assets/a9223934-7a91-4724-839f-f3112e72e67c" />

Bukama:
This would it looks like with your proposal:

<img width="2110" height="289" alt="Image" src="https://github.com/user-attachments/assets/2ca43b6b-aa7e-400e-98f7-785de6c78163" />

I think it will need a little tweak with rowspan / colspan. We should take a closer look when/after the page is converted to markdown (which is an continuously ongoing process #869)

PR #1362, opened 2025-08-05
Clarify Dependency Scope table

brkeh:
As mentioned in issue #1357, can we please clarify the dependency scope table on [Introduction to the Dependency Mechanism](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope) ?

Can we label the column and row headers? It took me a while to understand the table from the sentences above the table and I think header labels would it make clearer and easier to understand. My understanding is each row header in left column is a Main project dependency and each column header in first row is a Dependency project dependency. I wasn't able to get two header rows in Mark Down so I used HTML.

Following this checklist to help us incorporate your contribution quickly and easily:

- [ ] Each commit in the pull request should have a meaningful subject line and body.
- [ ] Run `mvn site` and examine output in `target/site` directory.

### DIFF
--- a/content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
+++ b/content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
@@ -95,13 +95,51 @@ There are 6 scopes:
 
 Each of the scopes (except for `import`) affects transitive dependencies in different ways, as is demonstrated in the table below. If a dependency is set to the scope in the left column, a transitive dependency of that dependency with the scope across the top row results in a dependency in the main project with the scope listed at the intersection. If no scope is listed, it means the dependency is omitted.
 
-|          |            |          |          |      |
-|:---------|:-----------|:---------|:---------|:-----|
-|          | compile    | provided | runtime  | test |
-| compile  | compile(*) | \-       | runtime  | \-   |
-| provided | provided   | \-       | provided | \-   |
-| runtime  | runtime    | \-       | runtime  | \-   |
-| test     | test       | \-       | test     | \-   |
+<table>
+  <thead>
+    <tr>
+      <th>Main project dependency</th>
+      <th colspan="4">Dependency project dependency</th>
+    </tr>
+    <tr>
+      <th></th>
+      <th>compile</th>
+      <th>provided</th>
+      <th>runtime</th>
+      <th>test</th>
+    </tr>
+  </thead>
+  <tbody>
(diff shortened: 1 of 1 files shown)


content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
```diff
@@ -95,13 +95,51 @@ There are 6 scopes:
...
 Each of the scopes (except for `import`) affects transitive dependencies in different ways, as is demonstrated in the table below. If a dependency is set to the scope in the left column, a transitive dependency of that dependency with the scope across the top row results in a dependency in the main project with the scope listed at the intersection. If no scope is listed, it means the dependency is omitted.
 
-|          |            |          |          |      |
-|:---------|:-----------|:---------|:---------|:-----|
-|          | compile    | provided | runtime  | test |
-| compile  | compile(*) | \-       | runtime  | \-   |
-| provided | provided   | \-       | provided | \-   |
-| runtime  | runtime    | \-       | runtime  | \-   |
-| test     | test       | \-       | test     | \-   |
+<table>
```
elharo, line 98: Is this just changing to HTML?

brkeh: No, click on the button to display the rich diff to see the change. My idea was to add header labels to the table. I wasn't able to get two header rows in Mark Down so I used HTML. My understanding is each row header in left column is a Main project dependency and each column header in first row is a Dependency project dependency.
image

elharo (changes requested):
I don't find the new headers clarifying. In particular, "Dependency project dependency" is awkward and unclear. Possibly different language might help here, but as is I prefer the original table.

elharo:
Not sure, but I do note that these cells and columns do not contain dependencies. They contain scopes.

Bukama (changes requested):
Sorry for my late comment: I was some weeks off and only did minor stuff (if even).

First, thank you for your PR!
According the change I'm with @elharo, the table is about dependency scopes. I could image that

* changing the "headers" to "project dependency scope" and "transitive dependency scope"
* moving the "left header" to a rowspan coloum (and maybe change the text by 90*, not sure by heart if CSS supports such things) makes it more clearer to which it belong
* adding sentence with an example 

might help

Bukama:
As this PR got abandoned, I took over and created #1411.

PR #1368, opened 2025-08-09
equals --> equal

elharo:
grammar

### DIFF
--- a/content/apt/pom.apt.vm
+++ b/content/apt/pom.apt.vm
@@ -416,7 +416,7 @@ mvn install:install-file -Dfile=non-maven-proj.jar -DgroupId=some.group -Dartifa
   Then the two sequences are compared token by token from beginning to end
   (left-to-right in the original strings). If each token 
   in one sequence compares equal to the corresponding token in the other sequence, then the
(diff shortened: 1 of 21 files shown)


Bukama:
Spotless ;)

Bukama:
ah I see its the `pom.apt.vm` I'm currently transforming this into markdown. Let me fix the error on my local feature branch.

Bukama:
Changed in my local branch

Bukama (changes requested):
I see 

* one "equals" fix
* one fix, where a wrong APT to MD conversion (`fals-->` instead of `false`)
* a lot of empty lines added

in addition I see #1370 where you remove a lot of empty lines.
Please do all formatting things in #1370 so this here is only about the "equals" as stated in title

elharo:
Yes, this is a mess. When I tried to do just the equals, fix spotless failed, apparently due to commit eeb1597ccc34ec61397a713e8cad537b65853798 which seems to have been committed directly without PR, CI, or code review. Applying spotless added all the other changes you see here. I'm trying to unravel this, but it's icky. This is why we need to protect the master branch so these mistakes can't happen.

PR #1370, opened 2025-08-10
Revert unreviewed commit

elharo:
The  change reverted seems to have broken spotless. We need to protect the master branch to prevent this sort of problem in the future.


Bukama:
Hey Elliote, what's the state of this draft?

elharo:
turned out to be a red herring, though we do need to protect the master branch to block stealth commits

PR #1371, opened 2025-08-11
spotless

elharo:
The change in content/markdown/guides/mini/guide-3rd-party-jars-remote.md looks like a bug in spotless. All changes in this PR were made by spotless


Bukama:
I can offer a review later today if you are finished with this draft. Otherwise I would not be able to for the next 1,5 weeks most probably

PR #1375, opened 2025-08-13
Maven Developer Centre --> Developing Maven

elharo:


sparsick:
I'm fine with the renaming. But we should ask more people for a review to avoid renaming ping pong.

Bukama:
Back from some week off / low activity:

I'm with Sandra to avoid naming ping pong and I also think we should take a look at all four "centres" if we decide to rename one or the other. As of now they may sound a bit weired, but the names are in line:

* User Centre
* Plugin Developer Centre
* Maven Repository Centre
* Maven Developer Centre

I suggest to open a discussion on ML or at least on Slack.


P.S. More and more I'm convinced that the whole page structure does not fit a good documentation site structure, but that's another thing.

raupachz:
+1

As mentioned before I think it is better to work on the documentation in small steps and save the brain juice for technical discussions. You are not going to confuse hundreds of people with this change overnight.

PR #1385, opened 2025-09-05
show how to document complex parameters

hboutemy:
need example of complex goal parameters documentation like https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#additionalDependencies
as plugin developer are often confused

=> improve https://maven.apache.org/guides/development/guide-plugin-documentation.html

while at it, simplification of the base project aspect, as a plugin is a project like any other: no need to give great examples of scm config, or issue tracking or anything not plugin-specific


content/apt/guides/development/guide-plugin-documentation.apt.vm
```diff
@@ -173,8 +110,8 @@ mvn site
 </project>
 +--------------+   
  
- The comments, annotations and plugin parameter 
- names are extracted from the plugin source and rendered in the Plugin Info page. In order for the generated site to 
+ The comments, annotations and plugin goals parameter 
+ names are extracted from the plugin source and rendered in the "Plugin Documentaiton" page. In order for the generated site to
```
nielsbasjes: Typo: Documentaiton --> Documentation

content/apt/guides/development/guide-plugin-documentation.apt.vm
```diff
@@ -207,7 +144,23 @@ public class ExampleMojo
 +--------------+   
 
  * the <<<@Component>>> and <<<readonly>>> parameters are not required to have any comments but it's still a good practice to provide one
- 
+
+ * if a parameter is not just a simple Java type or list of simple type, but a specific additional class, it is a good practice to give a sample
+   XML snippet showing the structure of objects that can be configured
```
nielsbasjes: Perhaps explicitly mention that fields in such an additional class should not be annotated with `@Parameter` because it is useless. And because of that do the XML snippet thing.

PR #1392, opened 2025-09-11
Require status checks - block direct push

slawekjaranowski:
- each commit must be done by PR to GitHub
- direct commit will be blocks to GitHub and to ASF gitbox
- build on GitHub must pass
- PR review is not required

As project is not releases should be not a problem to enable it.

It should protect us by broken master build

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -32,7 +32,11 @@ github:
     - FELIX
     - MRELEASE
   protected_branches:
-    master: {}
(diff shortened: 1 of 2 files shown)


slawekjaranowski:
@slachiewicz 
This change does not require review from others, only requires that change was build successful. 

I think that is reasonable for this project, when we direct push without checking we can break a site build.

PR #1409, opened 2025-10-02
update releases artifact, dependency, enforcer plugins

slawekjaranowski:


Bukama (approved):

slawekjaranowski:
There is no release notes in this project, so we not need labels 😄

PR #1410, opened 2025-10-02
Add hacktoberfest label to project

slawekjaranowski:


slawekjaranowski:
> Instead of a variety of different hackathon labels (hacktoberfest, hack-commit-push, etc.) I suggest to add and use "good first issue", "for beginners" or something like that.

This is a technical label - topic for repository 

we can add a labels for issues / PRs in https://github.com/apache/maven-gh-actions-shared/tree/main#labels-synchronization

PR #1412, opened 2025-10-03
[#1393] Add docs for Maven 4 behavioral change in phase execution

### LINKED ISSUE #1393: Improve docs on lifecycles and phases in Maven 4
### Improvement proposal

When running Maven with the syntax: `mvn [options] <phase>`

The expectation has been that only the explicitly specified phase (and its preceding phases) are executed. For example, when a plugin is bound to post-clean and the command `mvn clean` is run:

- Maven 3 behavior: The plugin does not run, since post-clean is defined after the clean phase in the clean lifecycle.
- Maven 4 behavior: The plugin does run, since Maven 4 executes the entire phase with `before:` and `after:` phases. 
 
Since Maven 4 now treats pre-* and post-* as before: and after:, this effectively means the clean lifecycle only contains the clean phase, with hooks provided by before: and after:, _which will always be run_.

I suggest we update https://maven.apache.org/whatsnewinmaven4.html to reflect this difference in behaviour.

Previous clean lifecycle should be also be updated at some point in that case probably: https://maven.apache.org/ref/3.9.9/maven-core/lifecycles.html#clean-lifecycle

Giovds:
Closes #1393

- [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/content/markdown/whatsnewinmaven4.md
+++ b/content/markdown/whatsnewinmaven4.md
@@ -419,8 +419,8 @@ Given the following plugin configuration, where a plugin is bound to the `post-c
 ```
 
 Running `mvn clean` using Maven 3 would NOT execute the plugin, because the build execution stops at the main phase, not executing it `post-` phase.
-To execute the plugin in Maven 3, the user has explicitly run `mvn post-clean` as a build goal.
-In Maven 4 running `mvn clean` will also execute the plugin, as both `before:` and `after:` phases are always executed.
+To execute the plugin in Maven 3, the user has to explicitly run `mvn post-clean` as a build goal.
+In Maven 4 running `mvn clean` will also execute the plugin, as both `before:` and `after:` phases are always executed and the `post-` phase is now an alias for `after:`.
 
 
 #### All- and each-phases


Bukama (commented):
Hi, thank you for the PR! I like the content you provide, but suggest to rewrite it a little bit to

* move some information,
* read less repeatitive, and
* using more active voice:

1) Alias
Move the alias information up to line 390:

old: 

> and had inconsistent naming, are deprecated.
> Don't use them!

new:

> and had inconsistent naming, are deprecated.
> They only act as aliases now and might be removed in the future.
> Don't use them!



2) Rewrite the new section a bit

--------

 
#### Behavioral change: Phase execution scope

Maven 4 introduces a behavioral change in how phases are executed.
For a given major phase, Maven 4 will aways execute **both** (`before:` and `after:`) minor phases of it.
This differs from Maven 3 where the `post-*` (now `after:`) phase is only executed if either the next major phase of the lifecycle is executed or the `post-*` phase is explicitly called.
 
 
The following example shows that this change might affect existing builds.
 
Given the following plugin configuration, where a plugin is bound to the `post-clean` phase.
 
```xml
<plugin>
  <groupId>org.example</groupId>
  <artifactId>some-plugin</artifactId>
  <executions>
    <execution>
      <!-- In Maven 3: Does NOT run when executing 'mvn clean', only with `mvn post-clean` -->
      <!-- In Maven 4: DOES run when executing 'mvn clean' -->
      <phase>post-clean</phase>
      <goals>
        <goal>cleanup</goal>
      </goals>
    </execution>
  </executions>
</plugin>
```

Running `mvn clean` using Maven 3 would **NOT** execute the plugin, because the build execution stops at the main phase, not executing it `post-` phase.
To execute the plugin in Maven 3, the user has explicitly run "mvn post-clean" as a build goal.
In Maven 4 it is enough to run `mvn clean` to execute the plugin.

---------------

3) Move the new section up after " Before- and after-phases, ordering of executions" (line 393) and before "All- and each-phases" as it belongs to the former one.

PR #1413, opened 2025-10-06
Refactor Update class for safer backpress and CI/CD stability

Manas-Dikshit:
This PR refactors the Update class to improve CI/CD stability and maintain the same update logic:

- Reuses a single Update instance instead of creating a new one per directory.
- Adds per-directory IOException handling to prevent the process from failing mid-way.
- Uses System.out.print for clearing console lines, avoiding extra newlines.

The update and lookup logic remain unchanged. These changes ensure smoother execution and better logging for multiple directory updates.

### DIFF
--- a/src/main/java/org/apache/maven/site/update/Update.java
+++ b/src/main/java/org/apache/maven/site/update/Update.java
@@ -34,12 +34,19 @@ import org.jsoup.nodes.Element;
  * effective site content.
  */
 public class Update {
-    public static void main(String[] args) throws IOException {
-        for (String dir : args) {
-            new Update().doUpdate(Paths.get("content/apt/" + dir + "/index.apt"));
-            System.out.println("\r\33[2K");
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/maven/site/update/Update.java
```diff
@@ -36,6 +36,17 @@
  * https://ci-maven.apache.org/job/Maven/job/maven-box/job/maven-dist-tool/job/master/site/dist-tool-check-errors.html
  */
 public class Update {
+    public static void main(String[] args) {
+    Update updater = new Update(); // reuse instance
```
elharo, line 40: I'm not sure this works. The value of the retired field might be different in a reused object. I'm not fully sure.

elharo: If one instance can be reused, then this should all just be static methods, no instances at all. @hboutemy

PR #1459, opened 2025-11-14
Add a page with logos

Bukama:
For an easier way to show what logos needs to be changed and to track the progress, I was asked to create a small page, showing all effected logos.

I also added a navigation entry for direct access. We can decide if we remove it after migration.


hboutemy (approved):
let's start like this
the menu entry is too visible IMHO, but not a blocker

PR #1473, opened 2025-11-30
Maven 3.9.12 - release notes

slawekjaranowski:

- [ ] Your pull request should address just one issue, without pulling in other changes.
- [ ] Each commit in the pull request should have a meaningful subject line and body.
- [ ] Run `mvn site` and examine output in `target/site` directory.

- [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

### DIFF
--- a/content/markdown/docs/history.md.vm
+++ b/content/markdown/docs/history.md.vm
@@ -106,7 +106,7 @@ $esc.h$esc.h Maven 3.9.x+
 <!--  release( date version announce latest-minor jdk rowspan ) -->
 <!-- The following two lines needs to be fixed in x days if the archive picks up the announcement mail. -->
 
-#release( "2025-12-xx" "3.9.12" "https://lists.apache.org/thread/TODO" "true" "Java 8" "13" )
+#release( "2025-12-13" "3.9.12" "https://lists.apache.org/thread/38xvc7olg6noop0ckr8m64s7s58v37vp" "true" "Java 8" "13" )
 #release( "2025-07-12" "3.9.11" "https://lists.apache.org/thread/0y3rkrszh7xckxjmvrst5dk1skzpfmm6" "" "" "" )
 #release( "2025-06-01" "3.9.10" "https://lists.apache.org/thread/bh8hr89omrbw19ychrklyh3rompn1hj7" "" "" "" )
 #release( "2024-08-17" "3.9.9" "https://lists.apache.org/thread/ov40dkfd1snyz2pdltsr5rzhdh2xd9yh" "" "" "" )


content/markdown/docs/3.9.12/release-notes.md
```diff
@@ -0,0 +1,81 @@
...
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Release Notes &#x2013; Maven 3.9.12
```
elharo: I think &#x2013; is wrong. This should be an em dash if it's a dash at all, though I suspect maybe a dash shouldn't be used here. DuckDuckGo LLM says"Use an em dash (—) to set off extra information or indicate a break in a sentence, while an en dash (–) is typically used to represent ranges, such as dates or numbers. The em dash is longer and can replace commas or parentheses, whereas the en dash is shorter and often signifies "to" in ranges."

content/markdown/docs/3.9.12/release-notes.md
```diff
@@ -0,0 +1,81 @@
...
+
+The Apache Maven team would like to announce the release of Maven 3.9.12.
+
+Maven 3.9.12 is [available for download][0].
+
+Maven is a software project management and comprehension tool. Based on the concept of a project object model (POM), Maven can manage a project's build, reporting, and documentation from a central place.
```
elharo: Let's remove this. I know we say it everywhere, but it's really techno-gibberish.

content/markdown/docs/3.9.12/release-notes.md
michael-o: It depends on the style, APA, Chicago, etc. There is no caonical reference. Don't change a single file without changing the rest.

michael-o (approved):

content/markdown/docs/3.9.12/release-notes.md
elharo: Requiring things to change everywhere is  common meta mistake, and I specifically disagree. It is better to do the wrong thing 19 times and the right thing once than doing the wrong thing 20 times. Nor is it required or even a good idea to fix the other 19 mistakes to avoid introducing the 20th here.

bmarwell: I agree with @elharo here. Let's start using the correct dashes from now an, does not hurt if old release notes are different.

slawekjaranowski: so which one will be correct here:
- en dash `–` (current)
- em dash `—`
- no dash

image

content/markdown/docs/3.9.12/release-notes.md
slawekjaranowski: done

content/markdown/docs/3.9.12/release-notes.md
michael-o: when using em dash then there are no spaces around.

michael-o: See https://www.merriam-webster.com/grammar/em-dash-en-dash-how-to-use

elharo: em dash or no dash, my preference reorder as

Maven 3.9.12 Release Notes

That feels like the more natural phrasing to me.

slawekjaranowski: dash removed 😄

michael-o: Use Door Dash now

PR #1477, opened 2025-12-06
Add note about non-flattening consumer POM by default

Bukama:


content/markdown/whatsnewinmaven4.md
```diff
@@ -81,6 +81,9 @@ The consumer POM is a **flattened** version of the build POM, meaning:
 This flattening ensures that consumers of your artifact have all the information they need without requiring access to
 parent POMs or understanding the internal structure of your multi-project build.
 
+The flatten feature is disabled by default to avoid unexpected behavior and better control about listing (transitive) dependencies in consumer POM.
+To enable a flattened consumer POM the `maven.consumer.pom.flatten` property must be set to `true`.
```
slawekjaranowski: We need add a property type -  system, user or project ...

gnodet: I think it's a user property.

slawekjaranowski: > I think it's a user property.

If it can not be a project property we need to clarify it

gnodet: I confirm, this is user properties only, not project.
See https://github.com/apache/maven/blob/master/impl/maven-core/src/main/java/org/apache/maven/internal/transformation/impl/DefaultConsumerPomBuilder.java
Those can be controlled using ${session.rootDirectory}/.mvn/maven-user.properties for a per-reactor configuration though.

Bukama: Fun that user properties are not described anywhere on the maven site (esp. not on the configuration page https://maven.apache.org/guides/mini/guide-configuring-maven.html )

Bukama: Sentence about type added

content/markdown/whatsnewinmaven4.md
```diff
@@ -67,7 +67,7 @@ article [Introduction to Maven Toolchains][8] by Maven maintainer Maarten Mulder
 
 ### Consumer POM
 
-Maven 4 generates a stripped down consumer POM that removes build information not needed by consumers, and deploys this
+Maven 4 can generate a stripped down consumer POM that removes build information not needed by consumers, and deploys this
 to the remote repository.
 It does not deploy the `pom.xml` used to build the project.
```
elharo, line 72: what happens when the consumer is building a dependent project?

content/markdown/whatsnewinmaven4.md
```diff
@@ -81,6 +81,10 @@ The consumer POM is a **flattened** version of the build POM, meaning:
 This flattening ensures that consumers of your artifact have all the information they need without requiring access to
 parent POMs or understanding the internal structure of your multi-project build.
 
+The flatten feature is disabled by default to avoid unexpected behavior and better control about listing (transitive) dependencies in consumer POM.
```
elharo: the consumer POM.
better control about  --> better control of

I'm also not sure that what's really going on here is control of listing the dependencies or control of the dependencies

content/markdown/whatsnewinmaven4.md
```diff
@@ -81,6 +81,10 @@ The consumer POM is a **flattened** version of the build POM, meaning:
...
+To enable a flattened consumer POM the user property `maven.consumer.pom.flatten` property must be set to `true`.
```
elharo: To enable a flattened consumer POM  --> To publish a flattened consumer POM instead of the full build POM,

content/markdown/whatsnewinmaven4.md
```diff
@@ -81,6 +81,10 @@ The consumer POM is a **flattened** version of the build POM, meaning:
...
+User properties can be controlled using `${session.rootDirectory}/.mvn/maven-user.properties` for a per-reactor configuration though.
```
elharo: delete though

content/markdown/whatsnewinmaven4.md
Bukama: Afaik that's why it was changed to opt-in configuration
@gnodet Can you please gives us/me more insight to phrase ist correctly in the article? Thank you!

gnodet: > what happens when the consumer is building a dependent project?

Not sure exactly what you mean, but if _anyone_ builds the project, it needs the source code (from the SCM repo or eventually a tarball), and the build POM will be there.   You cannot build a project from just the pom.xml uploaded on Maven Central.

gnodet: In addition, whenever the consumer POM is uploaded, the build POM is _also_ uploaded, but not as the _main_ POM, it has a `build` classifier.
See https://repo1.maven.org/maven2/org/jline/jline-terminal/3.30.6/ for example.

PR #1481, opened 2025-12-06
[MNGSITE-424] Update system dependency section

### LINKED ISSUE #854: [MNGSITE-424] How to replace system dependencies
**[Elliotte Rusty Harold](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=elharo)** opened **[MNGSITE-424](https://issues.apache.org/jira/browse/MNGSITE-424?redirect=false)** and commented

https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html says that system dependencies are deprecated. However it fails to suggest what developers should do instead.

There is an alternative for private jars not published to the Maven repos but it isn't documented anywhere. We should document it.  See https://stackoverflow.com/questions/4955635/how-to-add-local-jar-files-to-a-maven-project

I'm not sure what to do for dependencies that are expected to be in the JDK, e.g. srczip as used in Guava. Any suggestions?

some context: https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1539/
https://search.maven.org/artifact/com.google.guava/guava/29.0-android/bundle

---
No further details from [MNGSITE-424](https://issues.apache.org/jira/browse/MNGSITE-424?redirect=false)

Bukama:
closes #854

### DIFF
--- a/content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
+++ b/content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
@@ -788,9 +788,9 @@ The dependency is only looked up on this specific file path, which binds the bui
 The recommended approach is to upload the dependency to a [private hosted repository](/repository-management.html) accessible within the organization.
 This also allows differentiation between dependencies needed for compile/execution and those only needed for testing, by using _compile_ or _test_ scope.
 
-### Historical usage: Libraries of the JDK
+### Historical usage: Java EE (`javax`) libraries of the JDK
 
-In the past, the system scope was commonly used to tell Maven about dependencies provided by the JDK that were available as separate downloads earlier.
+In the past, the system scope was commonly used to tell Maven about Java EE (`javax` package) dependencies provided by the JDK that were available as separate downloads earlier.
 A typical examples is the Java Authentication and Authorization Service (JAAS):
 
 ```xml
@@ -808,5 +808,5 @@ A typical examples is the Java Authentication and Authorization Service (JAAS):
   ...
 </project>
 ```
-
-In general, those dependencies are available on Maven central nowadays.
+Most of those dependencies are available on Maven central nowadays.
+In general, you should never add any explicit dependencies for classes which are part of the JDK. 


content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
```diff
@@ -760,45 +760,53 @@ Starting from Maven 4.0, a new specific BOM packaging has been introduced. It al
 
 ## System Dependencies
 
-`Important note: This is deprecated.`
+`Important note: The usage of this scope is not recommended!`
 
-Dependencies with the scope _system_ are not looked up in the Maven repository system. Instead the `dependency` element contains a `systemPath` pointing to a jar on the local file system.
-
-The system scope is commonly used to tell Maven about dependencies provided by the JDK or the VM. System dependencies are especially useful for resolving dependencies on artifacts which are now provided by the JDK, but were available as separate downloads earlier. A typical examples is the Java Authentication and Authorization Service (JAAS):
+In rare occurrences it's needed to use a dependency which is not available in any repository, but only on local machine, for example a jar of some commercial application.
```
elharo: needed --> necessary
the local machine; For example,

content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
```diff
@@ -760,45 +760,53 @@ Starting from Maven 4.0, a new specific BOM packaging has been introduced. It al
...
 ```xml
-
 <project xmlns="http://maven.apache.org/POM/4.0.0">
   ...
   <dependencies>
     <dependency>
-      <groupId>javax.security</groupId>
-      <artifactId>jaas</artifactId>
-      <version>1.0.01</version>
+      <groupId>some.company</groupId>
```
elharo, line 775: what's wrong with the old example?

content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
```diff
@@ -760,45 +760,53 @@ Starting from Maven 4.0, a new specific BOM packaging has been introduced. It al
...
+      <systemPath>path/to/lib/the.jar</systemPath>
     </dependency>
   </dependencies>
   ...
 </project>
-
 ```
 
-If your artifact is provided by the JDK's `tools.jar`, the system path would be defined as follows:
+While the _system_ scope is technical possible, its usage is **not recommended**!
```
elharo: technical possible --> supported

content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
```diff
@@ -760,45 +760,53 @@ Starting from Maven 4.0, a new specific BOM packaging has been introduced. It al
...
+The dependency is only looked up on this specific file path, which extremely binds the build to individual machines.
```
elharo: delete extremely

content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
```diff
@@ -760,45 +760,53 @@ Starting from Maven 4.0, a new specific BOM packaging has been introduced. It al
...
+The recommended approach is to upload the dependency to a [private hosted repository](/repository-management.html) and share the access within the organization.
```
elharo: and share the access  --> accessible

content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
```diff
@@ -760,45 +760,53 @@ Starting from Maven 4.0, a new specific BOM packaging has been introduced. It al
...
+This also allows the differentiation between dependencies needed for compile/execution and those only needed for testing, by using _compile_ or _test_ scope.
```
elharo: allows the differentiation --> allows differentiation

content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
```diff
@@ -760,45 +760,53 @@ Starting from Maven 4.0, a new specific BOM packaging has been introduced. It al
...
+
+### Historical commonly usage: Libraries of the JDK
+
+In the past, the system scope was commonly used to tell Maven about dependencies provided by the JDK, but were available as separate downloads earlier.
```
elharo: , but were --> that were

content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
Bukama: I want to have difference examples for a the originally use case for this dependency scope ("commerical/non public") and the JDK use case (public, but not at Central), which was the most common use case in the past, but is not necessary anymore as they are in Central nowadays.

content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
```diff
@@ -760,45 +760,53 @@ Starting from Maven 4.0, a new specific BOM packaging has been introduced. It al
 
 ## System Dependencies
 
-`Important note: This is deprecated.`
+`Important note: The usage of this scope is not recommended!`
```
elharo: The usage of this scope --> Using this scope

content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
```diff
@@ -760,45 +760,53 @@ Starting from Maven 4.0, a new specific BOM packaging has been introduced. It al
...
-
 ```
 
-If your artifact is provided by the JDK's `tools.jar`, the system path would be defined as follows:
+While the _system_ scope is supported, its usage is **not recommended**!
+The dependency is only looked up on this specific file path, which binds the build to individual machines.
+The recommended approach is to upload the dependency to a [private hosted repository](/repository-management.html) accessible within the organization.
+This also allows differentiation between dependencies needed for compile/execution and those only needed for testing, by using _compile_ or _test_ scope.
+
+### Historical commonly usage: Libraries of the JDK
```
elharo: delete commonly

content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
```diff
@@ -760,45 +760,53 @@ Starting from Maven 4.0, a new specific BOM packaging has been introduced. It al
...
       <scope>system</scope>
-      <systemPath>${java.home}/../lib/tools.jar</systemPath>
+      <systemPath>${java.home}/lib/rt.jar</systemPath>
     </dependency>
   </dependencies>
   ...
 </project>
 ```
 
+In general, those dependencies are available on Maven central nowadays.
```
elharo: Is this true? If it's in rt.jar I wouldn't expect to find it in a different jar. That can be an error in JDK 9+ with JPMS.

elharo (changes requested):

content/markdown/guides/introduction/introduction-to-dependency-mechanism.md
Bukama: https://mvnrepository.com/artifact/javax.security/jaas/1.0.01

That's the page for the dependency of the example

elharo: https://mvnrepository.com/ is not maven central. It seems to me that if a dependency is in rt.jar (the core JDK class library) there should not be any dependency element for that in pom.xml, any more than we'd add one for java.util.HashMap

Bukama: Rephrased it a bit with focus on Java EE (javax). But I'm getting more and more to be point to delete the whole subsection at all

elharo (approved):

PR #1490, opened 2025-12-15
[MNGSITE-515] Update introduction to profiles

### LINKED ISSUE #873: [MNGSITE-515] Profiles not properly specified
**[Philippe Cloutier](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=chealer)** opened **[MNGSITE-515](https://issues.apache.org/jira/browse/MNGSITE-515?redirect=false)** and commented

I don't know if an unspecified feature is considered a bug or merely as an issue, but here goes anyway...

The POM Reference does not specify profiles beyond the following:

---

## Profiles

A new feature of the POM 4.0 is the ability of a project to change settings depending on the environment where it is being built. A `profile` element contains both an optional activation (a profile trigger) and the set of changes to be made to the POM if that profile has been activated. For example, a project built for a test environment may point to a different database than that of the final deployment. Or dependencies may be pulled from different repositories based upon the JDK version used. The elements of profiles are as follows:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
 ...

<profiles>
    <profile>
      <id>test</id>
      <activation>...</activation>
      <build>...</build>
      <modules>...</modules>
      <repositories>...</repositories>
      <pluginRepositories>...</pluginRepositories>
      <dependencies>...</dependencies>
      <reporting>...</reporting>
      <dependencyManagement>...</dependencyManagement>
      <distributionManagement>...</distributionManagement>
    </profile>
  </profiles>
</project>
```

...and the following section about activation.

Several sources including the following indicate that a _profile_ element can also contain a _properties_ element:
* https://www.baeldung.com/maven-profiles
* https://stackoverflow.com/questions/35468752/maven-profiles-with-variable-for-properties
* https://mkyong.com/maven/maven-profiles-example/

This also seems to contradict [the _Profiles in POMs_ section of _Introduction to Build Profiles_](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#profiles-in-poms), although I for one struggle to make sense of:

> `<properties>` (flag)(not actually available in the main POM, but used behind the scenes)(flag)

---

By the way, since profiles can be activated explicitly, the beginning of the _Activation_ subsection is misleading:

> The power of a profile comes from its ability to modify the basic POM only under certain circumstances. Those circumstances are specified via an activation element.

---
No further details from [MNGSITE-515](https://issues.apache.org/jira/browse/MNGSITE-515?redirect=false)

Bukama:
This PR removes almost the entire examples about `<profiles>` of the POM reference and integrates them into the profile guide to avoid duplication and inconsistent content.
Some missing or unclear parts were also added to the guide (as requested in the issue).

closes #873

### DIFF
--- a/content/markdown/guides/introduction/introduction-to-profiles.md
+++ b/content/markdown/guides/introduction/introduction-to-profiles.md
@@ -142,7 +142,7 @@ The following configuration will trigger the profile when the JDK's version _sta
 ```
 
 [Ranges](/enforcer/enforcer-rules/versionRanges.html) can also be used. Range values must start with either `[` or `(`.
-The following honours versions `1.3`, `1.4` and `1.5`.
+The following honours versions `1.3`, `1.4`, and `1.5`.
 
 ```xml
 <profiles>


content/markdown/pom.md
```diff
@@ -1980,15 +1979,15 @@ The elements of profiles are as follows:
...
     </profile>
   </profiles>
 </project>
 ```
 
 ### Activation
 
-Activations are the key of a profile. The power of a profile comes from its ability to modify the
-basic POM only under certain circumstances. Those circumstances are specified via an `activation` element.
+The `activation` element defines the conditions, when a profile becomes active.
```
Chealer: There should be no comma here.

This remains misleading, since profiles can be activated explicitly.

content/markdown/guides/introduction/introduction-to-profiles.md
```diff
@@ -106,7 +106,27 @@ This profile will automatically be active for all builds unless another profile
 Profiles can be automatically triggered based on the state of the build environment.
 These triggers are specified via an `<activation>` section in the profile.
 The implicit profile activation only refers to the container profile (and not to profiles in other modules with the same id).
-Here are some examples.
+
+_Note_: Before Maven 3.2.2 activation occurs when one or more of the specified criteria have been met. When the first positive result is encountered, processing stops and the profile is marked as active.
```
elharo: In 2025 I think we can stop talking about what happens pre-3.2.2

content/markdown/guides/introduction/introduction-to-profiles.md
```diff
@@ -106,7 +106,27 @@ This profile will automatically be active for all builds unless another profile
...
+Since Maven 3.2.2 activation occurs when all the specified criteria have been met.
+
+##### Active by default
+
+Boolean flag which determines if the profile is active by default. Is `false` by default.
+This flag is only evaluated if no other profile is explicitly activated via command line, `settings.xml` or implicitly activated through some other activator, otherwise it has no effect.
```
elharo: activator. Otherwise it has no effect.
or implicitly activated --> , or activated

content/markdown/guides/introduction/introduction-to-profiles.md
```diff
@@ -136,7 +157,24 @@ The following configuration will trigger the profile when the JDK's version _sta
 </profiles>
 ```
 
-_Note:_ an upper bound such as `,1.5]` is likely not to include most releases of 1.5, since they will have an additional "patch" release such as `_05` that is not taken into consideration in the above range.
+_Note:_ The value ranges match if the JDK version used for running Maven is between lower and upper bounds (either inclusive or exclusive).
```
elharo: the lower

content/markdown/guides/introduction/introduction-to-profiles.md
```diff
@@ -136,7 +157,24 @@ The following configuration will trigger the profile when the JDK's version _sta
...
+An upper bound such as `,1.5]` is likely not to include most releases of 1.5, since they will have an additional "patch" release such as `_05` that is not taken into consideration in the above range.
```
elharo: is likely not to include --> likely does not include

content/markdown/guides/introduction/introduction-to-profiles.md
```diff
@@ -136,7 +157,24 @@ The following configuration will trigger the profile when the JDK's version _sta
 </profiles>
 ```
 
-_Note:_ an upper bound such as `,1.5]` is likely not to include most releases of 1.5, since they will have an additional "patch" release such as `_05` that is not taken into consideration in the above range.
+_Note:_ The value ranges match if the JDK version used for running Maven is between lower and upper bounds (either inclusive or exclusive).
+An upper bound such as `,1.5]` is likely not to include most releases of 1.5, since they will have an additional "patch" release such as `_05` that is not taken into consideration in the above range. 
+
+If the range does not start with `[` or `(`, the value is interpreted as a (vendor) prefix.
```
elharo: do we need the parentheses around vendor?

content/markdown/guides/introduction/introduction-to-profiles.md
```diff
@@ -165,8 +203,16 @@ Each value can be prefixed with `!` to negate the matching. The values match if
 Since [Maven 3.9.7](https://issues.apache.org/jira/browse/MNG-5726) the value for `version` may be prefixed with `regex:`. In that case [regular pattern matching](https://docs.oracle.com/javase/tutorial/essential/regex/) is applied for the version matching and applied against the **lower case** `os.version` value.
 
 The actual OS values which need to match the given values are emitted when executing `mvn --version`.
+See the maven-enforcer-plugin's [Require OS Rule](/enforcer/enforcer-rules/requireOS.html) for more details about OS values.
```
elharo, line 204: might be better as an absolute URL in case this site is mirrored somewhere

content/markdown/guides/introduction/introduction-to-profiles.md
```diff
@@ -290,6 +336,9 @@ Since Maven 3.9.0 one can also evaluate the POM's packaging value by referencing
 
 ##### Files
 
+A given filename may activate the `profile` by the `existence` of a file, or if it is `missing`.
+**NOTE**: Interpolation for this element is limited to `${project.basedir}`, System properties and request properties.
```
elharo: comma in "properties, and"

content/markdown/guides/introduction/introduction-to-profiles.md
```diff
@@ -123,7 +141,8 @@ The following configuration will trigger the profile when the JDK's version _sta
 </profiles>
 ```
 
-[Ranges](/enforcer/enforcer-rules/versionRanges.html) can also be used. Range values must start with either `[` or `(`. Otherwise, the value is interpreted as a prefix. The following honours versions `1.3`, `1.4` and `1.5`.
+[Ranges](/enforcer/enforcer-rules/versionRanges.html) can also be used. Range values must start with either `[` or `(`.
+The following honours versions `1.3`, `1.4` and `1.5`.
```
elharo: Oxford comma after `1.4`

Bukama: Changed

PR #1494, opened 2025-12-16
Fixes :: #1489 :: Add security warnings against version ranges and special version qualifiers

### LINKED ISSUE #1489: Document and discourage min and max suffixes in version strings
### Bug description

This is a security vulnerability like other forms of version ranges

### Maven site URL where bug exists

https://maven.apache.org/pom.html#Version_Order_Specification

HarshMehta112:
Fixes : #1489

- [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/content/markdown/pom.md
+++ b/content/markdown/pom.md
@@ -314,6 +314,8 @@ Version requirements have the following syntax:
 Maven picks the highest version of each project that satisfies all the hard requirements of the dependencies on that project.
 If no version satisfies all the hard requirements, the build fails.
 
+**Security Warning**: The use of version ranges (hard requirements with brackets and parentheses) is **strongly discouraged** due to security concerns. Version ranges introduce unpredictability in builds, can lead to non-reproducible builds, and may unexpectedly pull in vulnerable versions of dependencies. Always prefer using explicit, fixed version numbers for dependencies to ensure build reproducibility and security. For similar reasons, the use of `min` and `max` suffixes in version strings should also be avoided (see [Version Order Specification](#version-order-specification) below).
+
 #### Version Order Specification
 
 If version strings are syntactically correct [Semantic Versioning 1.0.0](https://semver.org/spec/v1.0.0.html)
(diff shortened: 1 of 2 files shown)


content/markdown/security.md
```diff
@@ -31,6 +31,27 @@ the code, dependencies and repositories that are used in your build.
 If you want to use Maven to build untrusted code, it is up to you to
 provide the required isolation.
 
+## Security Best Practices
```
Bukama, line 34: I don't see the warning on this page. In my opinion this page is about CVE things.

content/markdown/pom.md
```diff
@@ -314,6 +314,8 @@ Version requirements have the following syntax:
 Maven picks the highest version of each project that satisfies all the hard requirements of the dependencies on that project.
 If no version satisfies all the hard requirements, the build fails.
 
+**Security Warning**: The use of version ranges (hard requirements with brackets and parentheses) is **strongly discouraged** due to security concerns. Version ranges introduce unpredictability in builds, can lead to non-reproducible builds, and may unexpectedly pull in vulnerable versions of dependencies. Always prefer using explicit, fixed version numbers for dependencies to ensure build reproducibility and security. For similar reasons, the use of `min` and `max` suffixes in version strings should also be avoided (see [Version Order Specification](#version-order-specification) below).
```
Bukama, line 317: I would like to only have warning a warning on this page that using those (as well as min/max, latest etc) are not recommended as they can lead to undesired behavior and pulled in artifacts and that we recommend using the enforcer-rule to enforce it. And then link to the page where version ranges are described, which is https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html and which also needs a warning with a more detailed explanation and examples and a link to the rule.

(at @elharo For me it's very similar to the system dependency: Maven support it and there are reasons to use it. Are those special and situations are rare? Yes, but they are there. So Maven should not warn about itself having a security problem, but explain it as  best practice to avoid this pitfall. It's not a security problem of Maven, but of every automatic dependency mechanism.

Bukama (changes requested):

content/markdown/pom.md
elharo: No,, this is much worse than system dependencies. It should never have been added and should never be used..

elharo: also, not all uses of brackets and parentheses introduce version ranges. They can also be used to pin a specific version during transitive dependency resolution

elharo (changes requested):

elharo (commented):
see https://jlbp.dev/JLBP-14

PR #1547, opened 2026-03-30
Add new apache-maven-project svg icons (for light and dark theme)

plumstone:
I recreated the apache-maven-project icon as svg in two versions, for light and dark theme. These can replace the old png and jpg versions of the icon in the long term.

I also changed the url on the icons to start with https (instead of http).

This change is closely related to my other PR https://github.com/apache/maven-fluido-skin/pull/325
It also considers the new ASF Logo https://github.com/apache/maven-site/issues/1396

Following this checklist to help us incorporate your
contribution quickly and easily:

- [x] Your pull request should address just one issue, without pulling in other changes.
- [x] Each commit in the pull request should have a meaningful subject line and body.
  Note that commits might be squashed by a maintainer on merge.
- [x] Run `mvn site` and examine output in `target/site` directory.
  Site will also be built on your pull request automatically and attached to GitHub Action result.

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

- [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Issue #1396, opened 2025-09-12
update site's logo with the new Oak Leaf ASF logo


hboutemy:
### Improvement proposal

context: https://news.apache.org/foundation/entry/introducing-the-asfs-new-logo

updating our website's top image is an easy first step before diving more
= https://github.com/apache/maven-site/blob/61fdaf2/content/resources/images/apache-maven-project.png
visible at the top of https://maven.apache.org/

slawekjaranowski:
for reference:
 - https://github.com/apache/maven-fluido-skin/issues/315

Bukama:
Thanks @slawekjaranowski  Will contact them!

Bukama:
As an information: No response to my mail so far.

Bukama:
Marketing asked me if we urgent need new logos, because the "Maven designer" is on vacation until new year. I answered that we don't and they don't have to try to find another designer.

michael-o:
I'd like to see the leaf left of the M as large the the M itself...

olamy:
1 then 2

chrisdutz:
I like the ones with more color 1 and 3.
Was there ever any special meaning to the "V"? Because it was higlited? Or was it just that it was located in the middle?

Bukama:
Vote for new logo: https://lists.apache.org/thread/dh51hyov2pgfk1xt1pkvp962kv6g5nfz

Bukama:
> Vote for new logo: https://lists.apache.org/thread/dh51hyov2pgfk1xt1pkvp962kv6g5nfz

Majority voted for Option 2 (https://lists.apache.org/thread/4vofz4bf8hgdkm1w50jxsjndf5jhphpp )

Bukama:
* Logos replaced in
  * site repository,
  * parent repository and 
  * maven 3.9.x branch (for 3.9.15).
* PR in fluido-skin repo opened
* Logo replaced in comdev SVN as described in docs. Some (they look somehow generated as they are not in the repository) images on project page are not recreated, but I guess the docs are outdated how often the page is recreated. So my task there is also done.

That's also logos I'm aware of and which were mentioned here. So this issue can be closed.

One more note: 
@hendrikebbers said he wants to reach out to a designer to convert the provided PNG to a SVG and then creates a PR with that.

PR #1551, opened 2026-04-03
Small showcase page

Bukama:


slawekjaranowski:
Are we not adding concepts 2 and 3 on purpose?

https://github.com/apache/maven-site/issues/1396

slawekjaranowski:
We also have a proposition https://github.com/apache/maven-site/pull/1547 for left logo

Bukama:
> Are we not adding concepts 2 and 3 on purpose?
> 
> #1396

These are concepts 1+2 which got most "likes" in the issue-thread

slawekjaranowski (approved):

Bukama:
> We also have a proposition #1547 for left logo

I see this slighty different as the current image with full name and URL also does not use the official logo (which can be seen on ASF project page and on top right of the Maven page)

PR #1554, opened 2026-04-09
Add new logo and apply it as banner

Bukama:
Notes:
* image-tag in site.xml offers width and height, but does not apply it, so I created a smaller image too with correct size
* The logo with the url is promoted in #1547


slawekjaranowski:
should be updated in parent also - https://github.com/apache/maven-parent/blob/master/src/site/site.xml

slawekjaranowski:
by the way we should check width and height in site.xml

slawekjaranowski (approved):

Bukama:
> should be updated in parent also - https://github.com/apache/maven-parent/blob/master/src/site/site.xml

Will do!

PR #1555, opened 2026-04-10
Add new banner for top left corner

Bukama:


content/site.xml
```diff
@@ -33,6 +33,10 @@ under the License.
     </fluidoSkin>
   </custom>
 
+  <bannerLeft href="https://www.apache.org/">
+    <image src="/images/apache-maven-project-light-theme.svg" width="489px" height="51px"/>
```
slawekjaranowski, line 37: width, height looks like working 😄

slawekjaranowski:
I'm not a graphics designer but once we use svg and once png it is ok?

slawekjaranowski:
new:

image

old

image

looks ok

slawekjaranowski (approved):

content/site.xml
Bukama: Yeah guess I had a bad day before

Bukama:
> I'm not a graphics designer but once we use svg and once png it is ok?

Me neither. Old logos/images are also a mix of both

slawekjaranowski:
> > I'm not a graphics designer but once we use svg and once png it is ok?
> 
> Me neither. Old logos/images are also a mix of both

so go will be better than now 😄 

we can change next time

PR #1564, opened 2026-04-18
Update Maven 3 xsd to reflect required elements

slachiewicz:
Regenerated with Modello 2.7 that contains https://github.com/codehaus-plexus/modello/pull/528 based on maven-3.10.x branch


slawekjaranowski:
Looks like copilot comments are right ...

We should review maven.mdo again ...

gnodet:
Agreed, those changes look wrong. Are they caused by modello upgrade? Is that a regression or a "bug fix"?

slawekjaranowski:
> Agreed, those changes look wrong. Are they caused by modello upgrade? Is that a regression or a "bug fix"?

fix in modello 2.6.0 
- https://github.com/codehaus-plexus/modello/pull/528

but show as the we required some of fields which should not be required ...

slawekjaranowski:
- https://github.com/apache/maven/issues/11979

slawekjaranowski:
next try with:
 - https://github.com/apache/maven/pull/11985

slawekjaranowski:
@Copilot check again

slawekjaranowski:
@copilot close - resolve conversations which was fixed

slawekjaranowski:
done in 
 - #1614

PR #1603, opened 2026-06-18
docs: remove SemVer 2.0.0 build-metadata examples from naming conventions guide

### LINKED ISSUE #1595: Incorrect version examples with SemVer v2 build metadata, which is not allowed in POM versions
### Bug description

In this section describing recommended version naming:

https://maven.apache.org/guides/mini/guide-naming-conventions.html#version-identifier

, the examples show build metadata:
```
// Build metadata
1.2.3+dfc0c87
2.3.4+15433
```
However, SemVer v1, which is explicitly referenced at the top of the section, does not define syntax build metadata.

Moreover, POM's detailed guide below explicitly calls out that the `+` character has no special meaning in POM.

https://maven.apache.org/pom.html#version-order-specification

> Important: This is only true for Semantic Versioning 1.0.0. The Maven version order algorithm is not compatible with Semantic Versioning 2.0.0. In particular, Maven does not special case the plus sign or consider build identifiers.

In addition to removing the incorrect examples, I would also suggest adding an explicit note about `+` not considered as a special character, so build metadata, if present, will be considered as a Maven qualifier. Might also be worth mentioning SemVer case sensitivity is ignored in Maven version comparisons.

### Maven site URL where bug exists

https://maven.apache.org/guides/mini/guide-naming-conventions.html#version-identifier

imharjot:
The naming-conventions guide recommends SemVer 1.0.0 but illustrated it with `+build` metadata, which is SemVer 2.0.0-only syntax that Maven does not special-case. Removes the incorrect examples, fixes the prose, and adds a clarifying note linking to the version-order spec.

Fixes #1595

### DIFF
--- a/content/markdown/guides/mini/guide-naming-conventions.md
+++ b/content/markdown/guides/mini/guide-naming-conventions.md
@@ -93,7 +93,7 @@ Examples
 3.5.42
 ```
 
-You can add labels for pre-releases or build metadata after the patch version.
+You can add labels for pre-releases after the patch version.
 Avoid using dates in those labels, because they are usually associated with unstable versions.
 
 Examples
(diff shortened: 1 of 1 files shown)


elharo:
This PR is misguided and misunderstands the current docs.

PR #1630, opened 2026-08-05
docs: correct non-existent bomClassifier element in What's new in Maven 4

### LINKED ISSUE #12661: Documented bomClassifier - No reference in XSD / Example description
### Affected version

Current WebSite

### Bug description

On the description page https://maven.apache.org/whatsnewinmaven4.html#new-packaging-type-bom 

> Therefore, the Maven team suggests that project BOMs should be generated as classified artifacts, using the `<bomClassifier>` element. This means that an imported BOM must not come from the same reactor as the current build but be available outside the project before the build. 

Here is given a reference to `bomClassifier`, but there is no reference in the XSD file for Model Version 4.1.0 nor have I found an example how to use `bomClassifier` 

Do I simply miss a thing here?

arimu1:
## Summary

Fixes [apache/maven#12661](https://github.com/apache/maven/issues/12661).

The [What's new in Maven 4](https://maven.apache.org/whatsnewinmaven4.html#new-packaging-type-bom) page claimed that project BOMs should be generated using a `<bomClassifier>` element. That element does not exist in the Maven 4.1.0 model or XSD, and there is no example of how to use it.

### Investigation

| Claim | Reality |
|-------|---------|
| `<bomClassifier>` POM element | **Does not exist** in `maven.mdo` / XSD 4.1.0 |
| Import BOM with classifier | **Supported** for model 4.1.0+ via standard dependency `<classifier>` (validator skips the "must be empty" rule when `is41OrBeyond`) |
| Generating classified BOMs | Done by attaching an additional POM artifact with a classifier (e.g. plugin config such as `bom-builder3`'s `bomClassifier` parameter) — not a model element |

This is a documentation correctness fix only (no model/XSD change).

### Changes

- Replace the false `<bomClassifier>` reference with the real import syntax using `<classifier>`
- Add a short import example
- Clarify that classified BOMs are attached artifacts, not a POM model element
- Keep the existing guidance about importing only external (non-reactor) BOMs

## Testing

- Docs-only change; no code build required
- Reviewed against `DefaultModelValidator` (classifier allowed for import scope when model ≥ 4.1.0) and confirmed no `bomClassifier` field in `api/maven-api-model`

## Checklist

- [x] Addresses a single issue without unrelated changes
- [x] Meaningful commit message

### DIFF
--- a/content/markdown/whatsnewinmaven4.md
+++ b/content/markdown/whatsnewinmaven4.md
@@ -170,9 +170,8 @@ standard `<classifier>` element on the import dependency:
 </dependencyManagement>
 ```
 
-There is no dedicated `<bomClassifier>` element in the POM model or XSD.
 Classified BOM artifacts are produced by attaching an additional POM artifact with a classifier (for example via a
-BOM-building plugin), not by a model element on the producing project.
+BOM-building plugin). Consumers import that artifact with the standard `<classifier>` element as shown above.
 
 Therefore, the Maven team suggests that project BOMs should be published as classified artifacts when appropriate.
 This means that an imported BOM must **not** come from the same reactor as the current build but be available outside
--- a/content/markdown/whatsnewinmaven4.md
+++ b/content/markdown/whatsnewinmaven4.md
@@ -152,9 +152,29 @@ the [live coding by Maven maintainer Karl Heinz Marbaise at IntelliJ IDEA Conf 2
 
 **Note**: With Maven 4, it's also possible to exclude dependencies that are declared by BOMs using the existing
 `<exclusions>` element.
-Also note that in Maven 4, importing BOMs with a classifier is now possible.
-Therefore, the Maven team suggests that project BOMs should be generated as classified artifacts, using the
-`<bomClassifier>` element.
+Also note that in Maven 4 (model version 4.1.0 and later), importing BOMs with a classifier is possible by using the
+standard `<classifier>` element on the import dependency:
+
+```xml
+<dependencyManagement>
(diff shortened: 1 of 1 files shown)


content/markdown/whatsnewinmaven4.md
```diff
@@ -152,9 +152,29 @@ the [live coding by Maven maintainer Karl Heinz Marbaise at IntelliJ IDEA Conf 2
...
+      <version>1.0.0</version>
+      <type>pom</type>
+      <classifier>bom</classifier>
+      <scope>import</scope>
+    </dependency>
+  </dependencies>
+</dependencyManagement>
+```
+
+There is no dedicated `<bomClassifier>` element in the POM model or XSD.
```
elharo: don't tell us what doesn't exist. There's an infinite number of things that don't exist

elharo (changes requested):

content/markdown/whatsnewinmaven4.md
arimu1: Thanks @elharo — good call.

Rewrote that note to state only the correct behavior: classified BOM artifacts are attached with a classifier (e.g. via a BOM-building plugin), and consumers import them with the standard `<classifier>` element as in the example above. Removed the “what doesn’t exist” phrasing.

arimu1:
Addressed @elharo’s review: the BOM note now describes only how classified BOMs are produced and imported via standard `<classifier>`, without saying what elements do not exist.

elharo (approved):

PR #1633, opened 2026-08-07
Lead with Markdown in the site and plugin documentation guides

slachiewicz:
Sits on top of #1632 — the first two commits here are that PR, since the plugin documentation guide has to be Markdown before its content can be rewritten. **Review the last commit only**; once #1632 merges this will rebase down to two files.

Two pieces of normative guidance still point new documentation at APT:

**`guides/mini/guide-site.md`** lists `apt` first and calls it the format "like this one" — which stopped being true when that guide itself became Markdown. It now lists `markdown` first, says plainly that new documentation should use it, and describes APT as a format that predates Maven's Markdown support: still supported, but the Maven project has moved its own documentation off it. Adds a pointer to `doxia-converter` for anyone converting an existing page.

**`guides/development/guide-plugin-documentation`** is the canonical instruction for documenting a plugin, and it gave a full APT template for `src/site/apt/index.apt` — so an author following it started in APT. It now shows the same page as `src/site/markdown/index.md`, with the YAML front matter that carries the title, author and date, plus a note about naming the file `index.md.vm` and keeping headings away from `##` if the page needs Velocity.

One thing worth seeing: that guide is itself a Velocity template, and Velocity reads `##` as a line comment **even inside a fenced code block**. The first version of this change silently lost `## Goals Overview` and `## Usage` from the rendered sample. It is now wrapped in `#[[ ]]#`. That only showed up because the built page was checked rather than the source.

Verified by building the site: the sample renders all its headings and no `#[[` markup leaks into the page.

### DIFF
--- a/content/markdown/guides/development/guide-plugin-documentation.md.vm
+++ b/content/markdown/guides/development/guide-plugin-documentation.md.vm
@@ -202,57 +202,61 @@ Below is the suggested site descriptor template (with maven-fluido-skin).
 
 - Introduction
 
-    The introduction is the front page of the plugin documentation. This is a good place to put any section and pages that need to be emphasized. The generated plugin parameter configuration should be linked here. Below is the suggested `src/site/apt/index.apt` template:
+    The introduction is the front page of the plugin documentation. This is a good place to put any section and pages that need to be emphasized. The generated plugin parameter configuration should be linked here. Below is the suggested `src/site/markdown/index.md` template:
 
+    ```markdown
+    #[[
+    ---
+    title: Introduction
+    author:
+      - Author
+    date: YYYY-MM-DD
+    ---
+
+    # Plugin Name
+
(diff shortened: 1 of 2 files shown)


content/markdown/guides/index.md.vm
```diff
@@ -0,0 +1,128 @@
...
+
+- [The Maven Community](../community.html)
+- [Helping with Maven](./development/guide-helping.html)
+- [Guide for New Committers](./mini/guide-new-committers.html)
+- [3rd Party Resources](../articles.html)
+
+Javadoc API
+-----------
+
+Here is some useful Javadoc API links to the current version of Maven:
```
elharo, line 118: is --> are

slachiewicz:
This is pure apt to Markdown conversion and now with Your previous commits I have conflicts

Issue #1395, opened 2025-09-12
Spotless destroys Yaml Front Matter metadata in markdown files


kwin:
### Bug description

We support both [YAML Front Matter](https://jekyllrb.com/docs/front-matter/) and [MultiMarkdown metadata](http://fletcher.github.io/MultiMarkdown-5/metadata.html) as metadata format for MD files (specifying title, author, ...). However the former is destroyed with spotless as it always replaces the tailing `---` by `-----------`

This regressed with https://github.com/apache/maven-site/commit/e89aac449589d2cf91550639a901d20e9d676066.

kwin:
This requires us to to disable spotless for now in maven-site. Therefore the linked/merged PR. However we can close as completed for now, as we implemented the workaround.

PR #1639, opened 2026-08-09
Add the cleanup steps that three retirements this year needed

slachiewicz:
Step 3 of the retirement plan was written for the repository alone and has drifted. Everything here came from actually retiring `maven-pdf-plugin`, `maven-stage-plugin` and `maven-verifier-plugin` — each gap below cost either a follow-up commit or left something wrong in public.

**New steps**

- **Ordering, stated up front.** Repository changes must merge before the INFRA archive request. It is the one hard constraint in the procedure and the page never said so; getting it wrong costs an unarchive/fix/re-archive round trip.
- **Delete the Actions run history and caches.** Removing `.github` stops new runs but leaves the old ones frozen in an archived repository — 1036 runs on stage, 787 on verifier, six caches each.
- **Delete stale topic branches.** Also frozen, and one still carrying a `Jenkinsfile` can keep a build job alive.
- **Replace the README**, which otherwise goes on explaining how to fork the repository and open pull requests forever.
- **Ask INFRA for the Jira project** in the same ticket as the repositories, as MPDF got under INFRA-26617.

**Corrections**

- The file is `Jenkinsfile`, not `.Jenkinsfile`, and CI has been `ci-maven.apache.org` for some time — the `builds.apache.org` link is dead.
- "Remove distribution" now says explicitly to remove the `.asc` and `.sha512` too. `maven-pdf-plugin` lost its zip in r75632 and kept both signature files, so downloads.apache.org served a PGP signature for a file that 404s — from March 2025 until this month.
- `maven-sources` needs the aggregator `<module>` removed alongside the manifest entry, or the aggregator build breaks for everyone who syncs it.

**The `.asf.yaml` step** gains the three settings that turned out to behave unintuitively. Each was verified by reading `infrastructure-asfyaml` rather than inferred, since its README documents none of it:

| Setting | Behaviour |
|---|---|
| `homepage` | `metadata.py` applies any truthy value but skips an absent or empty one, so it can be **set** but never **cleared**. Pointing it at maven.apache.org is therefore better than deleting the key. |
| `autolink_jira` | `autolink.py` only calls `create_autolink`; there is no deletion path, so removing the key leaves existing autolinks behind. |
| `features.issues` | `features.py` passes `has_issues=features.get("issues", False)`, so dropping the key while keeping the `features` block **disables issues** and hides every report the repository holds. Worth leaving enabled — an archived repo keeps its issues readable. |

`mvn spotless:check` passes.

<sub>Drafted with Claude — please verify</sub>

### DIFF
--- a/content/markdown/developers/retirement-plan-plugins.md
+++ b/content/markdown/developers/retirement-plan-plugins.md
@@ -100,18 +100,27 @@ If the vote passes, make one final release of the plugin (with its own standard 
 
 ## Step 3: Clean up after the release
 
-1. Remove the `.Jenkinsfile` from all branches. This will remove the project from https://builds.apache.org/job/maven-box/
+Everything that changes the repository has to be merged **before** INFRA archives it. An archived repository is read only, so anything left behind is frozen there permanently, and correcting it costs an unarchive/fix/re-archive round trip.
+
+1. Remove the `Jenkinsfile` from all branches. This will remove the project from the `maven-box` folder on https://ci-maven.apache.org/job/Maven/job/maven-box/
 2. Remove the `.github` directory from default branch. This will remove GitHub Actions
-3. Add " (RETIRED)" at the end of the project name in GitHub by updating `.asf.yaml`
-4. Ask INFRA to archive git repos (gitbox + github)
-5. When relevant update summary pages for [plugins](https://maven.apache.org/plugins/index.html) or [shared components](https://maven.apache.org/shared/index.html)
-6. Comment the [dist-tool configuration](https://ci-maven.apache.org/job/Maven/job/maven-box/job/maven-dist-tool/job/master/site/dist-tool.conf.html) entry. Also comment the plugin's entry in the `GetPrerequisites.java` and `ListBranchesReport` files.
-7. Remove from `Source Repository`
+   1. Deleting the directory stops new runs but leaves the old ones and their caches behind. Delete the workflow run history and the Actions caches as well, since they are frozen with the rest of the repository and serve no purpose once nothing can run again.
+3. Delete stale topic branches. Whatever survives is frozen by the archive, and a branch that still carries a `Jenkinsfile` can keep a build job alive.
+4. Update `.asf.yaml`:
+   1. Add " (RETIRED)" at the end of the project name.
+   2. Point `homepage` at https://maven.apache.org/ rather than the plugin's own site, which moves to the archive along with the plugin. Note that asfyaml only applies a homepage when one is given: removing the key stops it being reasserted but does not clear what GitHub already stores.
+   3. Remove `autolink_jira`. This stops the file claiming a Jira project that is being archived, though it does not delete autolinks GitHub has already created — ask INFRA for that.
+   4. Leave `features.issues` enabled. An archived repository keeps its issues readable, so disabling the feature would hide every report, including any migrated out of Jira.
+5. Replace the `README` with a short retirement notice. The file is otherwise a guide to forking the repository and opening pull requests, which is not actionable on a retired plugin. Keep it to what a visitor needs: that the plugin is retired, the last released version, and a link to the [mailing lists](https://maven.apache.org/mailing-lists.html) for questions. Do not link the plugin's own site, which will not outlive the plugin.
+6. Ask INFRA to archive git repos (gitbox + github). Ask in the same ticket for the plugin's Jira project to be archived, if it has one.
+7. When relevant update summary pages for [plugins](https://maven.apache.org/plugins/index.html) or [shared components](https://maven.apache.org/shared/index.html)
+8. Comment the [dist-tool configuration](https://ci-maven.apache.org/job/Maven/job/maven-box/job/maven-dist-tool/job/master/site/dist-tool.conf.html) entry. Also comment the plugin's entry in the `GetPrerequisites.java` and `ListBranchesReport` files.
+9. Remove from `Source Repository`
    1. [https://maven.apache.org/scm.html](https://maven.apache.org/scm.html) (includes the plantUML diagram)
-   2. [https://github.com/apache/maven-sources](https://github.com/apache/maven-sources)
-8. Remove distribution from current [dist area](https://dist.apache.org/repos/dist/release/maven/) (history remains available in [archive](https://archive.apache.org/dist/maven/)). This needs to be done by PMC members.
-9. Update board report. This needs to be done by PMC members.
-10. Announce the fact that the plugin has been retired/moved on the announce and users mailing lists. Explain to people what they should do if they would like to continue development of the plugin. Example for retirement email:
+   2. [https://github.com/apache/maven-sources](https://github.com/apache/maven-sources), moving the entry from `default.xml` to `retired.xml` and removing the matching `<module>` from the aggregator pom. Both, or the aggregator build breaks for everyone who syncs the manifest.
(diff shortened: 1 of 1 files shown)


content/markdown/developers/retirement-plan-plugins.md
```diff
@@ -100,18 +100,27 @@ If the vote passes, make one final release of the plugin (with its own standard
...
-6. Comment the [dist-tool configuration](https://ci-maven.apache.org/job/Maven/job/maven-box/job/maven-dist-tool/job/master/site/dist-tool.conf.html) entry. Also comment the plugin's entry in the `GetPrerequisites.java` and `ListBranchesReport` files.
-7. Remove from `Source Repository`
+   1. Deleting the directory stops new runs but leaves the old ones and their caches behind. Delete the workflow run history and the Actions caches as well, since they are frozen with the rest of the repository and serve no purpose once nothing can run again.
+3. Delete stale topic branches. Whatever survives is frozen by the archive, and a branch that still carries a `Jenkinsfile` can keep a build job alive.
+4. Update `.asf.yaml`:
+   1. Add " (RETIRED)" at the end of the project name.
+   2. Point `homepage` at https://maven.apache.org/ rather than the plugin's own site, which moves to the archive along with the plugin. Note that asfyaml only applies a homepage when one is given: removing the key stops it being reasserted but does not clear what GitHub already stores.
```
elharo: Note that asfyaml only applies a homepage when one is given. Removing the key stops it being reasserted but does not clear what GitHub already stores.

content/markdown/developers/retirement-plan-plugins.md
```diff
@@ -100,18 +100,27 @@ If the vote passes, make one final release of the plugin (with its own standard
...
+   3. Remove `autolink_jira`. This stops the file claiming a Jira project that is being archived, though it does not delete autolinks GitHub has already created — ask INFRA for that.
```
elharo: has already created. Ask INFRA for that.

content/markdown/developers/retirement-plan-plugins.md
```diff
@@ -100,18 +100,27 @@ If the vote passes, make one final release of the plugin (with its own standard
...
+   4. Leave `features.issues` enabled. An archived repository keeps its issues readable, so disabling the feature would hide every report, including any migrated out of Jira.
+5. Replace the `README` with a short retirement notice. The file is otherwise a guide to forking the repository and opening pull requests, which is not actionable on a retired plugin. Keep it to what a visitor needs: that the plugin is retired, the last released version, and a link to the [mailing lists](https://maven.apache.org/mailing-lists.html) for questions. Do not link the plugin's own site, which will not outlive the plugin.
```
elharo: "which will not outlive the plugin" That's unfortunate. We should try not to break links. Keeping old docs around is cheap, and code archaeology is a thing.

content/markdown/developers/retirement-plan-plugins.md
```diff
@@ -100,18 +100,27 @@ If the vote passes, make one final release of the plugin (with its own standard
...
+6. Ask INFRA to archive git repos (gitbox + github). Ask in the same ticket for the plugin's Jira project to be archived, if it has one.
+7. When relevant update summary pages for [plugins](https://maven.apache.org/plugins/index.html) or [shared components](https://maven.apache.org/shared/index.html)
```
elharo: When relevant,

content/markdown/developers/retirement-plan-plugins.md
```diff
@@ -100,18 +100,27 @@ If the vote passes, make one final release of the plugin (with its own standard
...
-8. Remove distribution from current [dist area](https://dist.apache.org/repos/dist/release/maven/) (history remains available in [archive](https://archive.apache.org/dist/maven/)). This needs to be done by PMC members.
-9. Update board report. This needs to be done by PMC members.
-10. Announce the fact that the plugin has been retired/moved on the announce and users mailing lists. Explain to people what they should do if they would like to continue development of the plugin. Example for retirement email:
+   2. [https://github.com/apache/maven-sources](https://github.com/apache/maven-sources), moving the entry from `default.xml` to `retired.xml` and removing the matching `<module>` from the aggregator pom. Both, or the aggregator build breaks for everyone who syncs the manifest.
+10. Remove distribution from current [dist area](https://dist.apache.org/repos/dist/release/maven/) (history remains available in [archive](https://archive.apache.org/dist/maven/)). This needs to be done by PMC members. Remove the `.asc` and `.sha512` files as well as the archive itself — leaving them behind publishes a signature for a file that no longer exists.
```
elharo: Itself — leaving them behind publishes a signature for a file that no longer exists.

-->

tself. Leaving them behind publishes a signature for a file that no longer exists.

elharo (changes requested):

content/markdown/developers/retirement-plan-plugins.md
slachiewicz: Is this comment or suggestion? If you suggest to change to other wording could you use GitHub option to propose changes, so merge would be easier

elharo: It's a suggestion. I.e. break run-on sentence into two. I'll try to propose edits but long lines make this more difficult for small changes

content/markdown/developers/retirement-plan-plugins.md
slachiewicz: Documentation stays on original location but is simply outdated - like for m-stage-p it's build on top of very very old skin that doesn't look too good now

slachiewicz:
Thanks for the review — all five comments are addressed in the latest commit.

The one about the plugin site was the important one, and you were right. I checked the history before changing it: **no Maven plugin retirement has ever removed a live site page.** All twelve retired plugins still have their directory under `website/components/plugins/` and all twelve return HTTP 200, going back to `maven-one-plugin` and `maven-idea-plugin` in 2013. `plugins-archives/` (539 directories) is filled by the *release* process, one per released version, so nothing "moves" at retirement time. I had written the guidance on the opposite assumption. It now says to leave `.asf.yaml` `homepage` pointing at the plugin site, and the README step keeps a link to it.

Also split the three run-on sentences and added the missing comma.

## On long lines — findings, and a question

You mentioned long lines make small suggestions hard. That is worth fixing, but I do not think tooling can do it. What I found:

**Spotless does not touch Markdown here at all.** `maven-parent` 49 configures only `<java>` and `<pom>`; this repo's own `<markdown><flexmark/>` block is commented out in `pom.xml`, citing #1395 (YAML front matter destroyed). That issue is closed, and current Spotless has a `YamlFrontMatter` extension, so the comment looks stale — re-enabling is probably possible now.

**The result is no convention at all.** Across the 248 files in `content/markdown`:

| | |
|---|---|
| average non-empty line | 65 chars |
| lines over 120 chars | 2959 |
| lines over 200 chars | 1344 |
| longest line | 1368 chars |

**The catch.** Spotless can enforce a width, via `<formatterOptions><rightMargin>120</rightMargin></formatterOptions>` — but `rightMargin` *reflows to fill* the margin. That makes it actively hostile to the thing that would actually help review: one sentence per line. Change a sentence, get a one-line diff, and your "propose changes" button works on it alone. Column wrapping does the reverse — fix one word mid-paragraph and everything after it reflows, so a typo shows up as a five-line diff.

So the two options are mutually exclusive, and no formatter can express "one sentence per line":

1. **Semantic line breaks as a convention**, applied only to text someone is already editing. No tooling, no mass reflow, no `git blame` damage.
2. **Enable Spotless with `rightMargin`**, mechanically enforced, at the cost of one 248-file reformat and giving up per-sentence diffs.

I lean towards 1, but you have reviewed far more of this repo than I have — which would you rather have? Happy to write it up as a short contributor-docs section and reflow Step 3 here as the worked example, in a separate PR so the guideline can be judged on its own.

<sub>Drafted with Claude — please verify</sub>

Issue #1645, opened 2026-08-19
Move the Doxia site into maven-site


slachiewicz:
`maven-doxia-site` is a separate repository, a separate Jenkins job and a separate svn
publication tree, for 21 pages that live at `/doxia/` on maven.apache.org. Folding it into
`maven-site` removes one repo, one job and one svnwcsub mount, and no URL has to change.

This issue is the plan and the tracking point. It is not a decision — three things at the
bottom need dev@ input first.

## How the site is published today

Both repos have the same shape: a single-module site build, a Jenkins job on the
`svn-websites` agent running `mvn site-deploy`, `maven-site-plugin` with `skipDeploy`, and
`maven-scm-publish-plugin` doing the svn commit. `maven-site` publishes to
`maven/website/content/`, `maven-doxia-site` to `maven/doxia/website/content/`.

The webserver working copy at `/www/maven.apache.org/content` is assembled by `svnwcsub`
from four checkouts, as documented in
[`developers/website/index.md`](https://maven.apache.org/developers/website/index.html):

| path | svn source |
|---|---|
| `/` | `maven/website/content/` |
| `/components` | `maven/website/components/` |
| `/doxia` | `maven/doxia/website/content/` |
| `/doxia/components` | `maven/doxia/website/components/` |

Released component documentation is never in git — release managers `svnmucc` it into the
`components/` trees, and the site build recreates symlinks into them from the
`components.links` files via `maven-antrun-plugin`. That task already globs
`**/components.links`, so it picks up nested ones without a change.

That `/doxia` mount is what makes this a two-party change: a `maven-site` PR **and** an
INFRA ticket, in that order.

## Content move

```
maven-doxia-site/content/markdown/**  ->  maven-site/content/markdown/doxia/**
maven-doxia-site/content/resources/** ->  maven-site/content/resources/doxia/**
```

Directory structure is preserved verbatim, so every relative link inside the Doxia pages
(`../macros/index.html`, `../doxia/doxia-modules/...`, `../doxia-archives/...`) keeps
resolving, and root-absolute links (`/xsd/...`, `/plugin-developers/...`) are unaffected.
`components.links` targets are relative, so the symlinks still land on `/doxia/components/...`.

`pom.xml`, `Jenkinsfile`, `.asf.yaml`, `README.md`, `LICENSE` and `NOTICE` are not moved.

Two lines in `maven-site/pom.xml`:

[...]

PR #1646, opened 2026-08-19
Move the Doxia site content into maven-site

slachiewicz:
Moves the `/doxia/**` pages out of `apache/maven-doxia-site` and into this repo, so the whole
of maven.apache.org is built from one source tree. Plan and open questions in #1645.

No `/doxia/` URL changes. Merging this is safe on its own: `svnwcsub` still mounts `/doxia`
from `maven/doxia/website/content/`, which shadows the `doxia/` directory this build starts
committing to `maven/website/content/`. The INFRA ticket that removes that mount is drafted in
#1645 and only goes in after this has merged and deployed.

`content/markdown/**` and `content/resources/**` from maven-doxia-site land under
`content/{markdown,resources}/doxia/` with their structure preserved, so relative links and
the `components.links` symlink targets keep resolving. `doxiaVersion` moves to `pom.xml`
because two `.vm` pages read it.

Three things do not survive the move:

- **Skin assets.** The skin writes its icons, fonts, `js/` and CSS into the *site root*, which
  is `/` here rather than `/doxia/`. The format table now loads `/images/icon_*_sml.gif`, and
  the two `*-archives/` rules in `/doxia/.htaccess` point at `/images/` and `/css/` so the
  archived release sites keep getting theirs.
- **Five project information pages.** `team`, `mailing-lists`, `ci-management` and
  `issue-management` redirect to the root equivalents; `project-info.html` redirects to
  `/doxia/`.
- **The site descriptor.** A build has one `site.xml`, so the second commit folds Doxia's
  menus into the Documentation menu next to Maven Tools. This is the part most worth
  reviewing — it is a judgement call, not a mechanical move, and question 1 in #1645.

Doxia's `site.css` turned out to be a strict subset of this repo's, so it is dropped rather
than duplicated.

If dev@ decides on the history graft (question 2 in #1645), this PR is superseded by that
merge rather than amended.

Verified: `mvn clean verify site` succeeds; `diff -rq` between the old standalone build and
`target/site/doxia` shows only the five report pages, `avatars/`, `css/`, `fonts/`, `img/`,
`js/` and the skin icons, all of which are served from the site root instead — and nothing
present in the new build that was not in the old. Resolving all 1554 local links in the built
Doxia pages leaves 36 dangling, down from 50: the remainder are the component symlinks and
`/download.cgi`, which exist only on the live server (200 today), and 14 links into the main
site that were broken in the standalone build and now resolve.

*This change was created with AI assistance.*


slachiewicz:
it's pure migration @elharo - when all pages would be visible (after INFRA work) we can do cleanups and deduplicate content

slachiewicz:
All three checks pass — `Verify and build Maven Site`, and both Jenkins jobs.

The INFRA side is [INFRA-28277](https://issues.apache.org/jira/browse/INFRA-28277). It asks INFRA to hold off until this has merged and deployed, so merging here is what unblocks it rather than the other way round.

*This comment was created with AI assistance.*

hboutemy (approved):
LGTM
and yes, prerequisite to https://issues.apache.org/jira/browse/INFRA-28277 
> Please remove *only* this entry from {{modules/svnwcsub/files/svnwcsub.conf}} in
infrastructure-puppet:
> * {{/www/maven.apache.org/content/doxia}} <- https://svn.apache.org/repos/asf/maven/doxia/website/content/

we can propose INFRA the PR

elharo:
"so the whole of maven.apache.org is built from one source tree" but that's not true. Pretty much all of https://maven.apache.org/plugins/ is not here for instance. 

I'm unconvinced that we want to have a mono-repo for the site without a monorepo for the code. Splitting plugin documentation from plugin code makes it impossible to have one PR that updates both.

In this case there's already a split repo, but perhaps this should merge into maven-doxia instead.

slachiewicz:
> "so the whole of maven.apache.org is built from one source tree" but that's not true. Pretty much all of https://maven.apache.org/plugins/ is not here for instance.
> 
> I'm unconvinced that we want to have a mono-repo for the site without a monorepo for the code. Splitting plugin documentation from plugin code makes it impossible to have one PR that updates both.
> 
> In this case there's already a split repo, but perhaps this should merge into maven-doxia instead.

yes, check my other proposal - to make maven.a.o as central place for documentaitons and not split it by each component. This way docs keept at plugin level will be smaller anyway and m.a.o more live - not depending on for this case Doxia releases happenings once per 2 years

hboutemy:
I just copied [`doxia/components`](https://svn.apache.org/repos/asf/maven/doxia/website/components/) to [`/components`](https://svn.apache.org/repos/asf/maven/website/components/)
then updated symlinks in e3d65984c04584e9b8d4a2ebcfb7ef8d2770ce0b

so we can completely drop in the end all the Doxia site specifics: this will really be something from the past

I'll update INFRA PR to drop `doxia/components` checkout...

I'm confident, we can merge this PR, no need to wait more, this simplification looks really reasonable and a good small step in the right direction (not always easy to find such a reasonable small step...)

PR #1665, opened 2026-09-02
Add Maven release metadata update workflow

slawekjaranowski:
Adds a weekly and manually triggered workflow that runs `mvn package -P update` and opens a pull request only if it changes tracked files.

The workflow skips execution while its prior update PR is still open.

- [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](https://www.apache.org/licenses/LICENSE-2.0).


slawekjaranowski:
It is accepted to use Maven from runner, the same is in verify-site.yml

slachiewicz (approved):

PR #1680, opened 2026-09-14
Remove Maven Shared IO from the shared index and SCM page

slachiewicz:
Maven Shared IO was retired by the [vote that passed on 2026-09-08](https://lists.apache.org/thread/8yv15hv6150vwt1qlgqvmpqxbf2jwpc0), and 3.1.0 is its final release. This removes its row from the shared index and from the SCM page, and takes its node out of the PlantUML diagram behind scm.html.

*This change was created with AI assistance.*


slawekjaranowski (approved):

Bukama (approved):
Note: Rebase needed due conflicts, but content change ok :)

